PR #7, opened 2022-12-31
Add ray trace and ray fan analyses

andibarg:
I followed the same structure as in the other analyses of this wrapper. The example is standalone and based on one of the examples provided by OpticStudio.

For the record, there seems to be multiple ways to run a sequential single ray trace with the API as described here: https://community.zemax.com/zos-api-12/single-ray-trace-in-the-zos-api-3051
I only implemented one of them.

Please review and let me know what you think.


LucVV:
Thank you for this contribution! I'll review and test it somewhere this week!

examples/Ray trace Double Gauss/Ray trace double gauss.py
```diff
@@ -0,0 +1,82 @@
...
+##########################
+## Connect to OpticStudio
+print('Connecting to Zemax API ...')
+zos = zp.ZOS()
+zos.wakeup()
+zos.create_new_application()
+oss = zos.get_primary_system()
+
+# Load example file
+file = "Double Gauss 28 degree field.zos"
```
LucVV, line 31:
```suggestion
file = "Double Gauss 28 degree field.zmx"
```

LucVV: @andibarg In my example folder, I have no *.zos* files, but only *.zmx* (or *.ZDA*). Is this also true for your folder, or did they update the format in newer OpticStudio versions?

zospy/analyses/raysandspots.py
```diff
@@ -0,0 +1,258 @@
...
+        section_start = inds[target]['rnum'] + 1  # + 1 to ignore section title
+
+        # Get endpoint
+        if order_of_appearance.index(target) == len(inds) - 1:
+            section_end = len(line_list)
+        else:
+            target2 = order_of_appearance[order_of_appearance.index(target) + 1]
+            section_end = inds[target2]['rnum']
+
+        # Read as dataframe
```
LucVV, line 55: @andibarg On certain machines, OpticStudio provides it's output with a comma rather than a point as decimal separator. In those cases, the numbers are strings rather than floats in the resulting dataframe. I think this can easily be resolved by adding ```.replace(',', '.')```. However, in some edge cases I ran into problems with this and I implemented some regex to find the comma delimiters (see [reports.py](https://github.com/MREYE-LUMC/ZOSPy/blob/c9490a82f31d247f3b4d2ec192c567680041d03a/zospy/analyses/reports.py)). 

Are you aware of any cases in which the text output would contain commas sthat are not intended to be decimal separators?

LucVV:
Hi @andibarg, 
I have reviewed the code and it looks very good. We are thinking about pulling it into a new version (v0.6.2) and then publishing that version. 

I have added two suggestions/questions about the code. Finally, I see that both ```ray_fan()``` and ```single_ray_trace()``` do not return the settings in the result. I will look into that somewhere in the coming week and adjust it such that it does.

examples/Ray trace Double Gauss/Ray trace double gauss.py
andibarg: @LucVV I am using OpticStudio 22.1.2 and for each example I have both _.zos_ and _.zmx_ files in the folder. So, things still work in my version if we change the filename extension to _.zmx_ and we make it backwork compatible, too.

zospy/analyses/raysandspots.py
andibarg: @LucVV Good point and yes. In the single ray trace text output, there could be commas in the comment column of the different surfaces. The ray fan text may also include comments.

Perhaps instead of the regex solution, we could make use of the decimal option in `pd.read_csv(Input, decimal=',')`. Apparently, the system default decimal can be found with `import locale` and `locale.localeconv()["decimal_point"]`. What do you think?

andibarg:
Hi @LucVV ,
I am happy to hear that. Thanks for your two suggestions, I just replied to them. Cheers.

LucVV:
Hi @andibarg 
Super! I hope to process everything the coming week!

PR #11, opened 2023-03-09
V0.6.2

LucVV:
v0.6.2 offers some new functionalities within [raysandspots.py](https://github.com/MREYE-LUMC/ZOSPy/blob/v0.6.2/zospy/analyses/raysandspots.py). Additionally, [an extra example](https://github.com/MREYE-LUMC/ZOSPy/tree/v0.6.2/examples/Ray%20trace%20Double%20Gauss) is added.


crnh:
@LucVV I ran the example and got the results displayed below; do they make sense (i.e. correspond with the results you got)?

image
image

LucVV:
@crnh Yes, these are the results I also got. The initial image shows that determination of the floating point separator works as expected!

crnh:
Great! Then I'll merge this one. In one of the next versions, we might want to use the decimal separator determination in the other analyses as well.

PR #13, opened 2023-05-05
Version 1.0.0

crnh:
Changes from GitLab for version 1.0.0


jwmbeenakker (approved):
As these changes have already been validated within the LUMC by members of dr. Beenakker's research group, only a quick review was performed, which showed that the current stable version of the internal gitlab repository is correctly pushed to the public GitHub repository.

PR #14, opened 2023-05-05
Add polarization analyses

andibarg:
I added two analyses from the polarization group, namely pupil map and transmission. I provided an example for both.

I also suggest to include a 'process_oncomplete' function in 'base.py', since this code appears at the end of every analysis. It could also be made into a decorator function. Perhaps a similar solution could be useful for creating and deleting the "txtoutfile" and "cfgoutfile"?

Please review and let me know what you think.


andibarg:
Oh, bad timing. I see you just released a new version with big changes. Let me know if we can work with my merge as is or if I should rewrite and pull request again another time.

crnh:
Hi, thanks for your PR! I think your analysis doesn't need much changes, but the analysis setup needs to be changed a bit.

There are a few other things that need to be changed:

- `process_oncomplete` is already part of the `Analysis` class and can therefore be removed;
- Please add type annotations to your functions;
- It would be really nice if you would contribute unit tests for your analyses (see https://github.com/MREYE-LUMC/ZOSPy/tree/main/tests);
- We now prefer Jupyter Notebooks instead of Python scripts for examples, since they allow to explain the code and show example output on GitHub, without having to run the code.

I'll try to have a better look next week and add comments to indicate which parts of your analyses need to be changed.

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+                         RawTextData=line_list, CgfOutFile=cfgoutfile, TxtOutFile=txtoutfile)
+
+    # cleanup if needed
+    if cleancfg:
+        os.remove(cfgoutfile)
+    if cleantxt:
+        os.remove(txtoutfile)
+
+    # Process oncomplete
+    process_oncomplete(analysis=analysis, oncomplete=oncomplete, ret=ret)
```
crnh, line 174: This is now handled by `analysis.complete(oncomplete, result)`

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+    settings.loc['Y-Phase'] = y_phase
+    settings.loc['Wavelength'] = int(wavelength)
+    settings.loc['Field'] = int(field)
+    settings.loc['Surface'] = surface
+    settings.loc['Sampling'] = sampling
+    settings.loc['Add Configs'] = add_configs
+    settings.loc['Sub Configs'] = sub_configs
+
+    # Create output
+    ret = AnalysisResult(analysistype=analysistype, data=data, settings=settings, metadata=metadata,
```
crnh, line 163: Do not abbreviate this variable name (or any other name)

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+
+    # Process oncomplete
+    process_oncomplete(analysis=analysis, oncomplete=oncomplete, ret=ret)
+
+    return ret
+
+
+
+
+def transmission(oss, sampling='32x32', unpolarized=False, jx=1, jy=0, x_phase=0, y_phase=0,
```
crnh, line 181: All comments from the previous analysis apply for this one as well

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+import numpy as np
+import pandas as pd
+
+from zospy import utils
+from zospy.analyses.base import AnalysisResult, AttrDict, process_oncomplete
+from zospy.api import constants
+import zospy.api.config as _config
+
+
+def polarization_pupil_map(oss, jx=1, jy=0, x_phase=0, y_phase=0, wavelength=1, field=1,
```
crnh, line 17: Add type hints and use the `zospy.analyses.base.OnComplete` Enum for `OnComplete` (see other analyses for example usages)

zospy/analyses/base.py
```diff
@@ -2,6 +2,18 @@
 from collections.abc import MutableMapping
 
 
+def process_oncomplete(analysis, oncomplete, ret):
```
crnh, line 5: This functionality is already part of the `Analysis` class

examples/polarization_prism_example.py
```diff
@@ -0,0 +1,72 @@
...
+# Connect to OpticStudio
+print('Connecting to Zemax API ...')
+zos = zp.ZOS()
+zos.wakeup()
+zos.create_new_application()
+oss = zos.get_primary_system()
+
+# Load example file
+file = "Prism using total internal reflection.zmx"
+testFile = zos.Application.SamplesDir + "\\Sequential\\Tilted systems & prisms\\" + file
```
crnh, line 31: Use snake_case names

examples/polarization_prism_example.py
```diff
@@ -0,0 +1,72 @@
...
+print('Loading file "%s" ...' %file)
+oss.load(testFile)
+
+# Polarization
+print('Calculating Transmission ...\n')
+ret0 = zp.analyses.transmission(oss, jx=jx, jy=jy,
```
crnh, line 37: I think it is not necessary to abbreviate this name

crnh (changes requested):

crnh:
Hi @andibarg, is there any progress on this PR? Feel free to contact us if you need more help with adapting your code to the changes made in v1.0.0!

andibarg:
> Hi @andibarg, is there any progress on this PR? Feel free to contact us if you need more help with adapting your code to the changes made in v1.0.0!

Hi @crnh, sorry for the delay. I will get back to it and adapt the code on the weekend.

zospy/analyses/base.py
andibarg: Yes. But sorry, I am not sure how I can remove this file change from the pull request. Can you do it somehow?

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+        cleantxt = False
+
+    analysis = oss.Analyses.New_Analysis_SettingsFirst(constants.Analysis.AnalysisIDM.loc[analysistype])
+
+    # Modify the settings file
+    an_sett = analysis.GetSettings()
+    an_sett.SaveTo(cfgoutfile)
+
+    # MODIFYSETTINGS are defined in the ZPL help files: The Programming Tab > About the ZPL > Keywords
+    an_sett.ModifySettings(cfgoutfile, 'PPM_JX', str(jx))
```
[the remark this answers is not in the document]
andibarg: The settings are unfortunately not accessible via the API, so I had to work with the settings file instead.

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,344 @@
...
+
+    # Run analysis
+    analysis.ApplyAndWaitForCompletion()
+
+    # Get results
+    analysis.Results.GetTextFile(txtoutfile)
+    line_list = [line for line in open(txtoutfile, 'r', encoding='utf-16-le')]
+
+    # Create output dict
+    data = AttrDict()
```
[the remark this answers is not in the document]
andibarg: Could you give me an example on how you implemented dataclasses instead of AttrDict? I could not find any in the other files. It seems that in some cases AttrDict is still used.

zospy/analyses/polarization.py
andibarg: Understood. Note that here modifying the settings file was very cumbersome. I think Zemax forgot to implement some functions so I had to reverse-engineer the byte arrays to figure out how to correctly change them. But I tested it and it works well now.

andibarg:
Hi @crnh. 

I implemented the changes. I may have made a bit of a mess because I edited the changes in the browser, directly on github. Sorry, in case I clicked on the wrong thing accidentally :).

Anyways, I tested the code earlier and it worked fine with the new version. Indeed, there were not too many big changes needed.

I usually do not work with Jupyter Notebooks so I did not convert my example for polarization analysis yet. Though since the example is not very interesting, we could also just remove it all together. It's up to you.

As I wrote in the comments above, I am not sure how on github I can remove the change for base.py without deleting the file in the PR. I hope you can just regard the changes in that file?

Thanks for your patience and best regards!

zospy/analyses/polarization.py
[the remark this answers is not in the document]
crnh: You're right, we'll convert the existing analyses later on. The idea is quite simple: define a dataclass with the same attributes that are used in the `AttrDict`. This may be less "smart" than the `AttrDict`, but at least you know what you get without having to interactively inspect the results, and [`AttrDict`s can cause some nasty problems](https://stackoverflow.com/questions/4984647/accessing-dict-keys-like-an-attribute).

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,381 @@
...
+        ind += 5
+        if 'Field Pos :' in line:
+            field = line.strip()
+            if not field in data:
+                data[field] = {}
+                ifield += 1
+                data[field]['Field number'] = ifield
+        elif 'Transmission at' in line or 'Total Transmission ' in line:
+            line_split = line.split(':')
+            data[field][re.sub(' +', ' ',line_split[0]).strip()] = float(line_split[-1])
```
crnh, line 331: @andibarg `field` may be undefined here

zospy/analyses/polarization.py
```diff
@@ -0,0 +1,381 @@
...
+            df = pd.read_csv(StringIO(''.join(line_list[ind:]).strip()),dtype=object,
+                             delimiter='\t', decimal=_config.DECIMAL)
+            df.columns = df.columns.str.strip()
+
+            # Find nan, truncate and add to data
+            nan_idxs = np.where(df['Tot. Tran'].isna())[0]
+            if len(nan_idxs) != 0:
+                first_nan = nan_idxs[0]
+                df = df.truncate(after=first_nan-1)
+            data[field][wvl] = df
```
crnh, line 345: @andibarg `wvl` and `field` may be undefined here

crnh:
@andibarg Unfortunately, I am unable to run your example. I suspect there's something wrong in the data parsing in `zospy.analyses.polarization.transmission`.

Full output:

```shell
Connecting to Zemax API ...
successfully unlocked Spatial components
Loading file Prism using total internal reflection.zmx ...
Calculating Transmission ...
Traceback (most recent call last):
  File "D:\code\zemax\zospy-github\examples\Polarization Prism\polarization_prism_example.py", line 33, in <module>
    return_transmission = zp.analyses.polarization.transmission(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\code\zemax\zospy-github\zospy\analyses\polarization.py", line 325, in transmission
    data[field][re.sub(" +", " ", line_split[0]).strip()] = float(line_split[-1])
                                                            ^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: '\t    1,000000000\n'

Process finished with exit code 1
```

Can you look into this? The polarization map part works fine.

zospy/analyses/polarization.py
crnh: Since I needed to debug this code due to the error mentioned below, I now get how this works. Due to the structure of the output, this will work.

crnh:
Hi @andibarg,

We looked into the problem with the `transmission` analysis ourselves and found the data parsing to be not very robust against small changes in the text output from OpticStudio. We suppose there are some small differences between your and our version of OpticStudio in the amount of whitespace used in the analysis output, which causes the parsing to fail.

We would like to merge your changes, but in its current state the `transmission` analysis is likely too fail in many cases, because it is not generic enough. We therefore propose to only support a single field in `transmission` for now. This makes the output much easier to parse, and parsing can then be performed in the same way as currently implemented in `polarization_pupil_map`. Later on, we might extend this to more complicated cases.

Note that `polarization_pupil_map` has been updated to use a `dataclass`. The result parsing was also slightly modified (it now uses regular expressions to find the right parameters in the text output), and this modification can easily be extended to the `transmission` analysis if we choose the temporary solution proposed above.

What are your thoughts on this?

andibarg:
Hi @crnh,

Yes, I was never too happy with the text output and the parsing for the transmission analysis either. But I tested it with multiple fields/settings and I saw no issue. It is unfortunate that OpticStudio does not provide more consistent text outputs - or in general, improve their API :).

I thought about trying the parsing with regular expressions at some point but it seemed even more cumbersome. Perhaps I could come back to that. But for the time being, we can do what you propose - support only a single field in the transmission analysis.

Thanks a lot for your time and work on the code.

crnh (approved):

andibarg:
@crnh. Looks great, thank you very much!

crnh:
You're welcome, and thanks for your contribution!

PR #17, opened 2023-05-08
Fix display problems with images in example notebook

crnh:
Apparently GitHub doesn't accept many different output formats for notebook outputs. Other viewers do accept the outputs from VS Code, but GitHub doesn't display them. After executing the notebooks with nbconvert, the results are correctly displayed.


jwmbeenakker (approved):
Thanks for fixing this!

PR #18, opened 2023-06-28
Adding function equivalent to ZPL numeric function OBJC($A) in zospy.functions.nce

Omnistic:
Hi Jan-Willem, and team,

This refers to [my answer in the Zemax Community](https://community.zemax.com/zos-api-12/is-there-a-zos-api-equivalent-to-the-zpl-objc-a-4170) about a ZOSAPI function equivalent to ZPL numeric function OBJC($A).

I somehow didn't manage to run ZOSPy with 2023 R1.03 though, so I couldn't test the example I wrote in the docstring.

I also don't know about unit testing (I'm not a software engineer) but I'm happy to help if you explain me how to do this kind of stuff.

Take care,

David

### DIFF
--- a/tests/functions/test_lde.py
+++ b/tests/functions/test_lde.py
@@ -3,27 +3,26 @@ import pytest
 from zospy.functions.lde import find_surface_by_comment
 
 
-class TestLDEFunctions:
-    @pytest.mark.parametrize(
-        "comment,case_sensitive,expected_indices",
-        [
-            ("Abcd", True, [0]),
-            ("Abcd", False, [0, 1, 2, 3]),
-            ("ABcd", True, [1]),
-            ("ABCD", True, []),
-            ("ABCD", False, [0, 1, 2, 3]),
-            ("efgh", True, []),
-            ("efgh", False, []),
-        ],
-    )
-    def test_can_find_surface_by_comment(self, simple_system, comment, case_sensitive, expected_indices):
-        simple_system.LDE.GetSurfaceAt(0).Comment = "Abcd"
-        simple_system.LDE.GetSurfaceAt(1).Comment = "ABcd"
-        simple_system.LDE.GetSurfaceAt(2).Comment = "ABCd"
-        simple_system.LDE.GetSurfaceAt(3).Comment = "ABCd"
+@pytest.mark.parametrize(
+    "comment,case_sensitive,expected_indices",
+    [
+        ("Abcd", True, [0]),
+        ("Abcd", False, [0, 1, 2, 3]),
+        ("ABcd", True, [1]),
+        ("ABCD", True, []),
+        ("ABCD", False, [0, 1, 2, 3]),
+        ("efgh", True, []),
+        ("efgh", False, []),
+    ],
+)
+def test_can_find_surface_by_comment(simple_system, comment, case_sensitive, expected_indices):
+    simple_system.LDE.GetSurfaceAt(0).Comment = "Abcd"
+    simple_system.LDE.GetSurfaceAt(1).Comment = "ABcd"
+    simple_system.LDE.GetSurfaceAt(2).Comment = "ABCd"
+    simple_system.LDE.GetSurfaceAt(3).Comment = "ABCd"
 
-        result = find_surface_by_comment(simple_system.LDE, comment=comment, case_sensitive=case_sensitive)
+    result = find_surface_by_comment(simple_system.LDE, comment=comment, case_sensitive=case_sensitive)
 
-        indices = [surface.RowIndex for surface in result]
+    indices = [surface.RowIndex for surface in result]
 
-        assert indices == expected_indices
+    assert indices == expected_indices
--- a/tests/functions/test_lde.py
+++ b/tests/functions/test_lde.py
@@ -6,7 +6,7 @@ from zospy.functions.lde import find_surface_by_comment
 @pytest.mark.parametrize(
     "comment,case_sensitive,expected_indices",
     [
-        ("Abcd", True, [0]),
+        ("Abcd", True, [0]), # When using .SurfaceNumber, the returned value matches the LDE numbering (starting at 0)
         ("Abcd", False, [0, 1, 2, 3]),
         ("ABcd", True, [1]),
         ("ABCD", True, []),
@@ -23,6 +23,6 @@ def test_can_find_surface_by_comment(simple_system, comment, case_sensitive, exp
 
     result = find_surface_by_comment(simple_system.LDE, comment=comment, case_sensitive=case_sensitive)
 
-    indices = [surface.RowIndex for surface in result]
+    indices = [surface.SurfaceNumber for surface in result]
 
     assert indices == expected_indices
--- a/tests/functions/test_lde.py
+++ b/tests/functions/test_lde.py
@@ -1,0 +1,29 @@
+import pytest
+
+from zospy.functions.lde import find_surface_by_comment
+
+
+class TestLDEFunctions:
+    @pytest.mark.parametrize(
+        "comment,case_sensitive,expected_indices",
+        [
+            ("Abcd", True, [0]),
+            ("Abcd", False, [0, 1, 2, 3]),
+            ("ABcd", True, [1]),
+            ("ABCD", True, []),
+            ("ABCD", False, [0, 1, 2, 3]),
+            ("efgh", True, []),
+            ("efgh", False, []),
+        ],
+    )
+    def test_can_find_surface_by_comment(self, simple_system, comment, case_sensitive, expected_indices):
+        simple_system.LDE.GetSurfaceAt(0).Comment = "Abcd"
+        simple_system.LDE.GetSurfaceAt(1).Comment = "ABcd"
+        simple_system.LDE.GetSurfaceAt(2).Comment = "ABCd"
+        simple_system.LDE.GetSurfaceAt(3).Comment = "ABCd"
+
+        result = find_surface_by_comment(simple_system.LDE, comment=comment, case_sensitive=case_sensitive)
+
+        indices = [surface.RowIndex for surface in result]
+
+        assert indices == expected_indices
--- a/tests/functions/test_lde.py
+++ b/tests/functions/test_lde.py
@@ -6,7 +6,7 @@ from zospy.functions.lde import find_surface_by_comment
 @pytest.mark.parametrize(
     "comment,case_sensitive,expected_indices",
     [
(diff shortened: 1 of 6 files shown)


jwmbeenakker:
Thanks for you contribution @Omnistic. We are about to push some new functionality to `ZOSPy`, so this one can swiftly be included the new versions.

Regarding `ZOSPy` not working with your OpticStudio, we are happy to help/fix the problem. But maybe then it is better to report an issue, so we can follow up (as this pull request will probably soon closed).

jwmbeenakker:
@crnh Would be willing to implement the unit test? I think the example David gave is already a sufficient test, which can be applied to the `simple_system`.

crnh (commented):
Hi @Omnistic, nice work! Two questions regarding the implementation:

1. The function currently returns -1 if no objects are found. This probably reflects the behaviour of the macro, but I think that's not required. What about returning `None` or an empty list? (The latter is common practice in libraries like NumPy and Pandas; check e.g. numpy.where or pandas.DataFrame.query).
2. Instead of a list of `int`s, you can also return a list of objects, since you're probably going to use the indices to get the objects anyways.

@LucVV suggested that this function can also be implemented for the LDE. I still need to investigate if we can do this using a shared implementation, or if we need completely separate implementations for each editor.

crnh:
> @crnh Would be willing to implement the unit test? I think the example David gave is already a sufficient test, which can be applied to the `simple_system`. 

Sure, unless @Omnistic is very eager to learn about unit testing... I think you did already give some hints!

zospy/functions/nce.py
```diff
@@ -33,3 +32,73 @@ def object_change_type(obj: _ZOSAPI.Editors.NCE.INCERow, new_type: constants.Edi
     # Apply
     new_surface_type_settings = obj.GetObjectTypeSettings(new_type)
     obj.ChangeType(new_surface_type_settings)
+
+def find_object_comment(nce: _ZOSAPI.Editors.NCE, comment: str, case_sensitive: bool=False) -> list[int] | int:
```
crnh, line 36: Could you change this to `find_object_by_comment`?

zospy/functions/nce.py
```diff
@@ -33,3 +32,73 @@ def object_change_type(obj: _ZOSAPI.Editors.NCE.INCERow, new_type: constants.Edi
...
+    """Function equivalent to ZPL numeric function OBJC($A).
```
crnh, line 37: Please do not refer to ZPL macros here, give a description of the function behaviour instead

zospy/functions/nce.py
```diff
@@ -33,3 +32,73 @@ def object_change_type(obj: _ZOSAPI.Editors.NCE.INCERow, new_type: constants.Edi
...
+        The Non-sequential Component Editor (NCE).
+    comment: str
+        String that is searched for in the Comment column of the NCE.
+    case_sensitive: bool=False
+        Flag that specifies whether the search is case-sensitive or not.
+
+
+    Returns
+    ----------
+    A list of integer indices corresponding to the rows of the NCE
```
crnh, line 51: Specify the type of the returned value here, in the same format as used elsewhere (i.e. [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#returns))

crnh (changes requested):

LucVV:
> Hi @Omnistic, nice work! Two questions regarding the implementation:
> 
> 1. The function currently returns -1 if no objects are found. This probably reflects the behaviour of the macro, but I think that's not required. What about returning `None` or an empty list? (The latter is common practice in libraries like NumPy and Pandas; check e.g. numpy.where or pandas.DataFrame.query).
> 2. Instead of a list of `int`s, you can also return a list of objects, since you're probably going to use the indices to get the objects anyways.

Regarding point 1, I think returning an empty list is preferable. In most cases, you will have to check the length of the returned list anyway befire you do that. Returning '[]' in stead of '-1' will save you from checking the tyoe of the return first.
Regarding point 2, I am also in favour of returning the object as you can get the position fron the object. We could also add a parameter to specify whether objects or object positions are returned if we want to support both possibilities.

> @LucVV suggested that this function can also be implemented for the LDE. I still need to investigate if we can do this using a shared implementation, or if we need completely separate implementations for each editor.

I have had a similar implementation in a project before to find the position of a surface in the LDE. I think you need to call slightly different API functions so separate implementations might be the easiest. I'll look into this.

crnh:
I agree with this. Although explicit type checking is not needed in this case (comparing a list with an int will return False, so that would be sufficient), it is still better to limit the amount of possible return types. An empty list therefore also has my preference.

Object indices can be easily found using a simple list comprehension, so I would leave this to the user. I don't think there are many cases in which you really need the indices instead of the objects.

As far as I know, the LDE rows and NCE objects implement the same interface. If the Comment property is part of this interface, we can probably use a shared implementation for both editors. Otherwise, it's indeed better to have two completely separate implementations.

crnh:
> As far as I know, the LDE rows and NCE objects implement the same interface. If the Comment property is part of this interface, we can probably use a shared implementation for both editors. Otherwise, it's indeed better to have two completely separate implementations.

I just checked this. `INCERow` and `ILDERow` do indeed both implement `IEditorRow`, but `Comment` is not a member of `IEditorRow`. So we need two separate implementations.

Omnistic:
Thank you for the comments. The function now returns a list of object that is empty is no match is found. As for the sequential mode implementation of this function, the only difference is that one has to call GetSurfaceAt from the LDE instead of GetObjectAt (from the NCE). Let me know if I should implement this in the same function or if you think  we need another one (happy to write it as well).

zospy/functions/nce.py
Omnistic: Done

Omnistic:
> > @crnh Would be willing to implement the unit test? I think the example David gave is already a sufficient test, which can be applied to the `simple_system`.
> 
> Sure, unless @Omnistic is very eager to learn about unit testing... I think you did already give some hints!

Not eager definitely, but if you have some pointers for me to have a rough idea of what it means, I'm all ears.

Omnistic:
Now that I'm able to better run ZOSPy, I noticed both function in nce.py wre refering to oss.NCE.InsertNewObjectAt(0) in their docstring. However, OpticStudio counts objects from 1. I think it should be replaced by oss.NCE.InsertNewObjectAt(1).

LucVV:
> However, OpticStudio counts objects from 1. I think it should be replaced by oss.NCE.InsertNewObjectAt(1).

I see, sorry the example is incorrect. I will adjust it soon! It seems to stem from the equivalent function for the LDE. Somehow,surface numbering starts at 0 in sequential mode but and at 1 in non-sequential mode. I have never fully figured out the reasoning behind that. I do think this is why two separate implementations of your function are required.


> now. Should object_change_type be called from zp somehow?


That is correct, `object_change_type()` is available by calling `zp.functions.nce.object_changetype()`. Or equivalently by importing it:

```python
from zospy.functions.nce import object_change_type()
object_change_type(...)
```
 Sorry that that is not clear in the example, I will update that.

LucVV:
Hi @Omnistic,

I have just added the same function for the LDE as `find_surface_by_comment()`. I have furthermore added a unit test. I think this unit test can swiftly be altered to work for the NCE. However, the `simple_system` that I use is sequential. We have a `nsc_simple_system` ready in the next release that might be optimal for this test. @crnh what do you think is the best way to go regarding to that non-sequential simple system?

zospy/functions/lde.py
```diff
@@ -71,3 +71,67 @@ def surface_change_type(surface: _ZOSAPI.Editors.LDE.ILDERow, new_type: constant
     # Apply
     new_surface_type_settings = surface.GetSurfaceTypeSettings(new_type)
     surface.ChangeType(new_surface_type_settings)
+
+
+def find_surface_by_comment(
+    lde: _ZOSAPI.Editors.LDE, comment: str, case_sensitive: bool = False
```
crnh, line 77: I would make `case_sensitive` `True` by default, so the function looks for exact matches unless specified otherwise. This also goes for the NCE version.

tests/functions/test_lde.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from zospy.functions.lde import find_surface_by_comment
+
+
+class TestLDEFunctions:
```
crnh, line 6: Imo a test class is not needed here (classes are purely used for grouping tests, and grouping at file level seems sufficient here).

zospy/functions/lde.py
```diff
@@ -71,3 +71,67 @@ def surface_change_type(surface: _ZOSAPI.Editors.LDE.ILDERow, new_type: constant
...
+    lde: ZOSAPI.Editors.LDE
+        The Lens Data Editor (LDE)
+    comment: str
+        String that is searched for in the Comment column of the NCE.
+    case_sensitive: bool=False
+        Flag that specifies whether the search is case-sensitive or not. Defaults to False.
+
+    Returns
+    -------
+    list[_ZOSAPI.Editors.LDE.ILDERow]
```
crnh, line 94: Please do not use the underscore when specifying ZOSAPI types in docstrings (the underscore only has a meaning in the context of "importing" the stub files). This also goes for the NCE function.

crnh (changes requested):

crnh:
> We have a `nsc_simple_system` ready in the next release that might be optimal for this test. @crnh what do you think is the best way to go regarding to that non-sequential simple system?

Introducing the `nsc_simple_system` in this PR as well seems to me the easiest way to add proper unit tests for `zospy.functions.nce.find_objects_by_comment`. Otherwise we won't be able to test the unit tests...

tests/functions/test_lde.py
LucVV: I have reverted it to a single function

zospy/functions/lde.py
LucVV: I agree with this and have altered it in `find_surface_by_comment()`

LucVV:
I have added the non-sequential simple system for testing purposes.

zospy/functions/lde.py
LucVV: Updated in the LDE function!

jwmbeenakker:
@crnh Could you "quickly" add the unit test for the non-sequential part, so we can include it in the merge with all our internal stuff for version 1.1.0

zospy/functions/lde.py
Omnistic: Done

zospy/functions/lde.py
Omnistic: Done in the NCE one as well

crnh (approved):

PR #22, opened 2023-07-04
Add actions for linting and formatting

crnh:
Automatically run the format and lint tox environments on each push and PR.
This doesn't update the code, but indicates if contributions meet our standards.

It would be nice to run the unit tests as well, but this is not feasible, due to our dependency on OpticStudio.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,0 +1,20 @@
+name: Run tox environments
+
+on:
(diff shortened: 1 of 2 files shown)


jwmbeenakker (approved):
Good suggestion to automatically perform these steps for each push/merge

LucVV (approved):
Looks good!

PR #26, opened 2023-07-17
Initialize with fewer lines

andibarg:
I implemented changes discussed in https://github.com/MREYE-LUMC/ZOSPy/discussions/20. Indeed, I can now initialize ZOSpy with just two lines:

```python
import zospy as zp

zos = zp.ZOS()
oss = zos.connect_as_standalone()
``` 

Note that I introduced a new function `zos.connect_as_standalone()` which is identical to `zos.create_new_application()`, because I find that here the word 'standalone' is important to be more consistent with the ZOS API.

At the moment, I don't see anything breaking and it should be backward compatible, as long as ZOSpy is initialized exactly as described in the readme (without another assert in the main script to check connection). I have tested it with my own scripts but I have not run a unit test, since I am currently trying to wrap my head around that :).

Let me know what you think.

Edit: Corrected code snippet.

### DIFF
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -466,14 +466,7 @@ class ZOS:
             return self.get_primary_system()
             
         return True
-            logger.critical("OpticStudio Licence is not valid for API, connection not established")
-            raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
-            return False
-        else:
-            if not return_primary_system:
-                return True
-            else:
-                return self.get_primary_system()
+
 
     def create_new_application(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
         """Creates a standalone Zemax Opticstudio instance.
@@ -500,13 +493,16 @@ class ZOS:
 
         if not self.Application.IsValidLicenseForAPI:
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
-            raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
+            
+            if return_primary_system:
+                raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
+            
             return False
-        else:
-            if not return_primary_system:
-                return True
-            else:
-                return self.get_primary_system()
+        
+        if return_primary_system:
+            return self.get_primary_system()
+            
+        return True
 
     def connect_as_standalone(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
         """Creates a standalone Zemax Opticstudio instance.
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -413,10 +413,12 @@ class ZOS:
         if not self.Application.IsValidLicenseForAPI:
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
             raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
-            logger.critical("OpticStudio Licence is not valid for API, connection not established")
-
-        
-        return self.get_primary_system()
+            return False
+        else:
+            if not return_primary_system:
+                return True
+            else
+                return self.get_primary_system()
     
     def create_new_application(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
         """Creates a standalone Zemax Opticstudio instance.
@@ -443,10 +445,13 @@ class ZOS:
 
         if not self.Application.IsValidLicenseForAPI:
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
-
-        assert self.Application.IsValidLicenseForAPI, "OpticStudio Licence is not valid for API, connection not established"
-        
-        return self.get_primary_system()
+            raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
+            return False
+        else:
+            if not return_primary_system:
+                return True
+            else
+                return self.get_primary_system()
  
     def connect_as_standalone(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
         """Creates a standalone Zemax Opticstudio instance.
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -296,7 +296,7 @@ class ZOS:
 
     _instances = set()
 
-    def __new__(cls, preload: bool = False, zosapi_nethelper: str = None, *args, **kwargs):
+    def __new__(cls, *args, **kwargs):
         if len(cls._instances) >= 1:
             # As the number of applications within runtime is limited to 1 by Zemax, it is logical to also limit the
             # number of ZOS instances
@@ -307,7 +307,7 @@ class ZOS:
                 "new one."
             )
 
-        instance = super(ZOS, cls).__new__(cls, *args, **kwargs)
+        instance = super(ZOS, cls).__new__(cls)
 
         return instance
 
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -318,6 +318,8 @@ class ZOS:
 
         logger.info("ZOS instance initialized")
 
+        self.wakeup()
+
     def wakeup(self, preload: bool = False, zosapi_nethelper: str = None):
         """Wakes the zosapi instance.
 
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -407,7 +407,6 @@ class ZOS:
             raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
 
-        assert self.Application.IsValidLicenseForAPI, "OpticStudio Licence is not valid for API, connection not established"
         
         return self.get_primary_system()
     
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -263,33 +263,56 @@ class ZOS:
     """A Communication instance for Zemax OpticStudio.
 
     This class can be used to establish a link between Python and Zemax OpticStudio through .NET,
-    and control OpticStudio. The connection is established as following:
-
-    Simple:
-    1. self.wakeup().
-    2. self.create_new_session() or self.connect_as_extension().
-    3. self.get_primary_system()
-
-    Advanced:
-    1. self._load_zos_dlls()
-    2. self._update_constants()
-    3. self.create_new_session() or self.connect_as_extension().
-    4. self.get_primary_system()
-
-    After connection, many OpticStudio functionalities are controllable through ZOS.ZOSAPI()
-
-    Attributes
-    ----------
-        ZOSAPI (None | netModuleObject): The ZOSAPI interface or if loaded.
-        ZOSAPI_NetHelper (None | netModuleObject): The ZOSAPI_NetHelper interface if loaded.
+    and subsequently control OpticStudio. The connection is established in two ways, the preferred method as wel as a
+    legacy method for backwards compatability. See examples.
 
     Parameters
     ----------
-    preload: bool
-        A boolean indicating if nested namespaces should be preloaded.
-    zosapi_nethelper: str, optional
-        Optional filepath to the ZOSAPI_NetHelper dll, if None, the windows registry will be used to find
-        ZOSAPI_NetHelper dll. Defaults to None.
+    preload : bool
+        A boolean indicating if nested namespaces should be preloaded upon initiating ZOS. Defaults to False.
+    zosapi_nethelper : str | None
+        Optional filepath to the ZOSAPI_NetHelper dll that is required to connect to OpticStudio. If None, the
+        Windows registry will be used to find the ZOSAPI_NetHelper dll. Defaults to None.
+
+
+    Attributes
+    ----------
+    ZOSAPI : None | netModuleObject
+        The ZOSAPI interface once loaded, else None.
+    ZOSAPI_NetHelper : None | netModuleObject
+        The ZOSAPI_NetHelper interface once loaded, else None.
+
+    Examples
+    --------
+    >>> # Preferred methods:
+    >>>
+    >>> # Connecting as extension:
+    >>> import zospy as zp
+    >>> zos = zp.ZOS()
+    >>> oss = zos.connect_as_extension(return_primary_system=True)
+    >>>
+    >>> # Launching OpticStudio in standalone mode:
+    >>> import zospy as zp
+    >>> zos = zp.ZOS()
+    >>> oss = zos.create_new_application(return_primary_system=True)
+    >>>
+    >>>
+    >>>
+    >>> # Legacy methods:
+    >>>
+    >>> # Connecting as extension:
+    >>> import zospy as zp
+    >>> zos = zp.ZOS()
+    >>> zos.wakeup()
+    >>> zos.connect_as_extension()
+    >>> oss = zos.get_primary_system()
+    >>>
+    >>> # Launching OpticStudio in standalone mode:
+    >>> import zospy as zp
+    >>> zos = zp.ZOS()
+    >>> zos.wakeup()
+    >>> zos.create_new_application()
+    >>> oss = zos.get_primary_system()
     """
 
     _OpticStudioSystem = OpticStudioSystem
@@ -312,7 +335,19 @@ class ZOS:
         return instance
 
     def __init__(self, preload: bool = False, zosapi_nethelper: str = None):
-        """Initiation of the ZOS Instance."""
+        """Initiation of the ZOS Instance.
+
+        The ZOS instance can subsequently be used to connect to OpticStudio. See the examples in the class docstring for
+        more information
+
+        Parameters
+        ----------
+        preload : bool
+            A boolean indicating if nested namespaces should be preloaded upon initiating ZOS. Defaults to False.
+        zosapi_nethelper : str | None
+            Optional filepath to the ZOSAPI_NetHelper dll that is required to connect to OpticStudio. If None, the
+            Windows registry will be used to find the ZOSAPI_NetHelper dll. Defaults to None.
+        """
         logger.debug("Initializing ZOS instance")
 
         self.ZOSAPI: _ZOSAPI = None
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -300,7 +300,7 @@ class ZOS:
         if len(cls._instances) >= 1:
             # As the number of applications within runtime is limited to 1 by Zemax, it is logical to also limit the
             # number of ZOS instances
-            raise ValueError(
+            raise ConnectionRefusedError(
                 "Cannot have more than one active ZOS instance.\n\n"
                 "Since OpticStudio limits the number of connections to the ZOS-API to 1, only a single ZOS instance "
                 "is allowed. Re-use the existing instance, or delete the existing instance prior to initializing a "
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -456,6 +456,17 @@ class ZOS:
 
         if not self.Application.IsValidLicenseForAPI:
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
+            
+            if return_primary_system:
+                raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
+            
+            return False
+        
+        if return_primary_system:
+            return self.get_primary_system()
+            
+        return True
+            logger.critical("OpticStudio Licence is not valid for API, connection not established")
             raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
             return False
         else:
--- a/zospy/zpcore.py
+++ b/zospy/zpcore.py
@@ -400,8 +400,8 @@ class ZOS:
         Returns
         -------
         bool | OpticStudioSystem
-            True if a valid connection is made, else False. If 'return_primary_system = True', the function
-            runs 'get_primary_system()' and directly returns OpticStudioSystem.
+            True if a valid connection is made, else False. If return_primary_system = True, the function
+            runs ZOS.get_primary_system() and directly returns OpticStudioSystem.
         """
         self._assign_connection()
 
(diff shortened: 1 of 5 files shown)


LucVV:
Thank you for this PR! We will look into it shortly.

With regards to your example code, I estimate that is should be:
```python
import zospy as zp

zos = zp.ZOS()
oss = zos.connect_as_standalone()
```
is that correct?

I think one of the main things that have to be evaluate is whether calling `wakeup()` twice when the current (4-line) initialization is used would brake something down the line. Else, we might want to implement a check within `wakeup()` that checks if it is already called (e.g. by checking whether `self.Connection` is `None` or adding a new private attribute to ZOS().

With regards to the additional function name, I personally like this but thus far we have kept everything in ZOSPy consistent with the API naming. I think it is good to wait for input of @crnh and/or @jwmbeenakker before accepting that.

Finally, with regards to the unit-testing (a bit off topic), there is a [README](https://github.com/MREYE-LUMC/ZOSPy/tree/main/tests) that should get you started on how to run them. Some IDEs also support them, but this depends on which IDE you use. To keep this PR focused on the PR, feel free to contact me via the [email listed in the README](https://github.com/MREYE-LUMC/ZOSPy#contact) for further support.

andibarg:
> Thank you for this PR! We will look into it shortly.
> 
> With regards to your example code, I estimate that is should be:
> 
> ```python
> import zospy as zp
> 
> zos = zp.ZOS()
> oss = zos.connect_as_standalone()
> ```
> 
> is that correct?
> 
> I think one of the main things that have to be evaluate is whether calling `wakeup()` twice when the current (4-line) initialization is used would brake something down the line. Else, we might want to implement a check within `wakeup()` that checks if it is already called (e.g. by checking whether `self.Connection` is `None` or adding a new private attribute to ZOS().
> 
> With regards to the additional function name, I personally like this but thus far we have kept everything in ZOSPy consistent with the API naming. I think it is good to wait for input of @crnh and/or @jwmbeenakker before accepting that.
> 
> Finally, with regards to the unit-testing (a bit off topic), there is a [README](https://github.com/MREYE-LUMC/ZOSPy/tree/main/tests) that should get you started on how to run them. Some IDEs also support them, but this depends on which IDE you use. To keep this PR focused on the PR, feel free to contact me via the [email listed in the README](https://github.com/MREYE-LUMC/ZOSPy#contact) for further support.

Yes, you are right about the example code.

I tried `wakeup()` multiple times in my scripts and did not see any errors. Does it not just (re-)load the dll?

crnh:
In addition to @LucVV's points, I would like to point out the following:

1. This change modifies the return type of `connect_as_extension` and `create_new_application`, which is by definition a breaking change. Raising an exception from a method that previously didn't is also a breaking change. If it is backwards compatible within certain limits, it is not backwards compatible.
2. [Do not use `assert` in production code](https://stackoverflow.com/questions/944592/best-practice-for-using-assert). `assert` statements will not be evaluated if the code is optimized. Instead, raise an exception. It is probably best to create our own exception classes (something like `OpticStudioConnectionError` would be a possibility here). Having our own exception classes could come in handy anyways; at least it is something I considered to add before.
3. `connect_as_standalone` is  a nice-to-have, but it is not in line with the corresponding function in the ZOS-API. Since ZOSPy's users are expected to understand the fundamentals of the ZOS-API, I don't think `create_new_application` is confusing, especially because this is also well-documented in our README. I am not against this change in case @jwmbeenakker and @LucVV like to include this alias, but in my opinion the added value is fairly minimal.

crnh:
Regarding point 2, Python's [ConnectionRefusedError](https://docs.python.org/3/library/exceptions.html#ConnectionRefusedError) seems appropriate in this case. We then don't need to add our own exception class for this.

andibarg:
Ok, I see there are many issues with my suggestions and perhaps it is not worth the effort to make these changes.

@crnh Thanks for your important points. Regarding 2: In my opinion, having just a boolean return to indicate (un-)successful connection, is not the best solution. Especially, if in the four lines of example code we do not suggest to check for connection errors or even indicate that it's possible/good practice to do so. Regarding 3: I see the term "Standalone Application" used by Zemax for this feature in the GUI and the programming examples all over the place. Then again, I do not have a problem either with `create_new_application` :)

Overall, it is up to you if you want to look further into the changes at all or leave zpcore as it is. Either way would be fine by me.

crnh:
> Ok, I see there are many issues with my suggestions and perhaps it is not worth the effort to make these changes.

I definitely do not oppose your changes, but am rather glad you want to contribute. Most of the suggested changes take 1 - 2 lines of code, so I'd say that's worth the effort 👍

> Regarding 2: In my opinion, having just a boolean return to indicate (un-)successful connection, is not the best solution.

I agree. My point was only to not use `assert`, but raise a `ConnectionRefusedError` instead. This is more clear and also prevents issues with the exception not being raised under certain conditions.

zospy/zpcore.py
```diff
@@ -318,6 +318,8 @@ def __init__(self):
 
         logger.info("ZOS instance initialized")
 
+        self.wakeup()
```
LucVV, line 321: Please make sure the arguments for wakeup are parsed. 

```suggestion
        self.wakeup(preload=preload, zosapi_nethelper=zoapi_nethelper)
```

To do so, they have to be made available in __init__. I think zosapi_nethelper requires to be an argument, as the possibility of passing the path to this nethelper is relevant. For preload, I think this is not required for most users and might be added as a kwarg. @crnh do you agree? If so, the code above can be changed to `preload=kwargs.get(preload, False)`

crnh: I personally do not really see the advantage of using `**kwargs` here as we don't deal with a large number of parameters, and `**kwargs` tends to make the interface more opaque.
I propose to add them both in the same way as is currently done in `ZOS.wakeup`.

zospy/zpcore.py
```diff
@@ -374,15 +376,15 @@ def _assign_connection(self):
         else:
             logger.debug("ZOSAPI_Connection() already assigned self.Connection")
 
-    def connect_as_extension(self, instancenumber: int = 0) -> bool:
-        """Creates a standalone Zemax Opticstudio instance.
+    def connect_as_extension(self, instancenumber: int = 0) -> OpticStudioSystem:
```
LucVV, line 379:
```suggestion
    def connect_as_extension(self, instancenumber: int = 0, return_system: bool = False) -> bool | OpticStudioSystem:
```

I have discussed this with @crnh and @jwmbeenakker. We are in favor of directly returning the opticstudiosystem, but fear that it might be a breaking change for some users. Therefore, we would like to implement it as a boolean argument for now. This way, the change is not a breaking change for now. When we approach a version that will induce breaking changes, we can change the default behaviour.

Please update the docsting and function return accordingly

LucVV: @crnh do we want this argument to be named `return_system` or `return_opticstudiosystem`, or differently?

zospy/zpcore.py
```diff
@@ -391,21 +393,22 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
 
         self.Application = self.Connection.ConnectAsExtension(instancenumber)
 
-        if self.Application.IsValidLicenseForAPI:
-            return True
-        else:
+        if not self.Application.IsValidLicenseForAPI:
```
LucVV, line 396:
```suggestion
        if not self.Application.IsValidLicenseForAPI:
            logger.critical("OpticStudio Licence is not valid for API, connection not established")
            raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
```

zospy/zpcore.py
```diff
@@ -391,21 +393,22 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
...
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
-            return False
 
-    def create_new_application(self) -> bool:
+        assert self.Application.IsValidLicenseForAPI, "OpticStudio Licence is not valid for API, connection not established"
```
LucVV, line 399:
```suggestion
```

zospy/zpcore.py
```diff
@@ -391,21 +393,22 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
...
+        
+        return self.get_primary_system()
```
LucVV, line 401: Please update based on the suggested function parameters

zospy/zpcore.py
```diff
@@ -391,21 +393,22 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
...
+    
+    def create_new_application(self) -> OpticStudioSystem:
```
LucVV, line 403: See suggestions at `connect_as_extension()`

zospy/zpcore.py
```diff
@@ -414,11 +417,24 @@ def create_new_application(self) -> bool:
...
-        else:
+        if not self.Application.IsValidLicenseForAPI:
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
-            return False
+
+        assert self.Application.IsValidLicenseForAPI, "OpticStudio Licence is not valid for API, connection not established"
+        
+        return self.get_primary_system()
+ 
+    def connect_as_standalone(self) -> OpticStudioSystem:
```
LucVV, line 427: See suggestions at `connect_as_extension()`

zospy/zpcore.py
```diff
@@ -414,11 +417,24 @@ def create_new_application(self) -> bool:
...
+        """Creates a standalone Zemax Opticstudio instance.
+
+        The application will be assigned to self.Application.
```
LucVV, line 430:
```suggestion
        Equal to self.create_new_application().
        
        The application will be assigned to self.Application.
```

I think it is important to acknowledge that the functions are exactly the same

crnh: Maybe it's better to refer to `ZOS.create_new_application()`, since `self` only has a meaning inside a class

andibarg:
> > > Ok, I see there are many issues with my suggestions and perhaps it is not worth the effort to make these changes.
> > 
> > 
> > I definitely do not oppose your changes, but am rather glad you want to contribute. Most of the suggested changes take 1 - 2 lines of code, so I'd say that's worth the effort 👍
> 
> Exactly, we are glad that you want to contribute! I'd also say that it is worth the effort! The reason we are treading carefully is that it affects lines of code that are used by every user, so any breaking changes might affect all.
> 
> I have made some suggestions to improve the code.
> 
> For now, we would like to add a parameter to `connect_as_extension()` and `connect_as_standalone()` that indicates the return, and keep the default as boolean to not introduce breaking changes. In future versions we can update this default parameter to the behavior that you suggest!

I understand that we have to be careful with these changes. And just for saving 2 lines of code - I did not want to insist :). But now I did the review and it seems to work well. Curious to hear what you think.

zospy/zpcore.py
```diff
@@ -282,13 +282,21 @@ class ZOS:
...
+        Optional filepath to the ZOSAPI_NetHelper dll, if None, the windows registry will be used to find
+        ZOSAPI_NetHelper dll. Defaults to None.
     """
 
     _OpticStudioSystem = OpticStudioSystem
 
     _instances = set()
 
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, preload: bool = False, zosapi_nethelper: str = None, *args, **kwargs):
```
crnh, line 299: These arguments are only needed in `__init__`, not in `__new__` (and are therefore not used in the body of this method). `__new__` is only used to create a new, uninitialized `ZOS` object. These parameters are only necessary during initialization of that object.

zospy/zpcore.py
```diff
@@ -374,15 +385,23 @@ def _assign_connection(self):
         else:
             logger.debug("ZOSAPI_Connection() already assigned self.Connection")
 
-    def connect_as_extension(self, instancenumber: int = 0) -> bool:
-        """Creates a standalone Zemax Opticstudio instance.
+    def connect_as_extension(self, instancenumber: int = 0, return_primary_system: bool = False) -> bool | OpticStudioSystem:
+        """Connects to Zemax Opticstudio as extension.
 
         The application will be assigned to self.Application.
```
crnh, line 391:
```suggestion
        The application will be assigned to ZOS.Application.
```

zospy/zpcore.py
```diff
@@ -391,21 +410,31 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
 
         self.Application = self.Connection.ConnectAsExtension(instancenumber)
 
-        if self.Application.IsValidLicenseForAPI:
-            return True
-        else:
+        if not self.Application.IsValidLicenseForAPI:
```
crnh, line 413: A `ConnectionRefusedError` is now also raised if `return_primary_system` is `False`, which introduces a breaking change. This also makes the return statement redundant, since it is unreachable.

```suggestion
        if not self.Application.IsValidLicenseForAPI:
            logger.critical("OpticStudio Licence is not valid for API, connection not established")
            
            if return_primary_system:
                raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
            
            return False
        
        if return_primary_system:
            return self.get_primary_system()
            
        return True
```

zospy/zpcore.py
```diff
@@ -414,11 +443,33 @@ def create_new_application(self) -> bool:
 
         self.Application = self.Connection.CreateNewApplication()
 
-        if self.Application.IsValidLicenseForAPI:
-            return True
-        else:
+        if not self.Application.IsValidLicenseForAPI:
```
crnh, line 446: Please apply the changes suggested for `connect_as_extension` here as well.

zospy/zpcore.py
```diff
@@ -374,15 +385,23 @@ def _assign_connection(self):
...
+        """Connects to Zemax Opticstudio as extension.
 
         The application will be assigned to self.Application.
 
+        Parameters
+        ----------
+        instancenumber: int
+            An integer to specify the number of instance used.
+        return_primary_system: bool
+            A boolean indicating if OpticStudioSystem should be returned directly.
```
crnh, line 398:
```suggestion
            A boolean indicating if the primary  OpticStudioSystem  should be returned.
```

zospy/zpcore.py
```diff
@@ -374,15 +385,23 @@ def _assign_connection(self):
...
+
         Returns
         -------
-        bool
-            True if a valid connection is made, else False
+        bool | OpticStudioSystem
+            True if a valid connection is made, else False. If return_primary_system = True, the function
```
crnh, line 403:
```suggestion
            `True` if a valid connection is made, else `False`. If `return_primary_system` is `True`, the function
            returns the primary `OpticStudioSystem`.
```

zospy/zpcore.py
```diff
@@ -391,21 +410,31 @@ def connect_as_extension(self, instancenumber: int = 0) -> bool:
...
             return False
-
-    def create_new_application(self) -> bool:
+        else:
+            if not return_primary_system:
+                return True
+            else:
+                return self.get_primary_system()
+    
+    def create_new_application(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
```
crnh, line 423: The suggestions for the docstring of `connect_as_extension` go for this one as well.

zospy/zpcore.py
```diff
@@ -414,11 +443,33 @@ def create_new_application(self) -> bool:
...
             logger.critical("OpticStudio Licence is not valid for API, connection not established")
+            raise ConnectionRefusedError("OpticStudio Licence is not valid for API, connection not established")
             return False
+        else:
+            if not return_primary_system:
+                return True
+            else:
+                return self.get_primary_system()
+ 
+    def connect_as_standalone(self, return_primary_system: bool = False) -> bool | OpticStudioSystem:
```
crnh, line 456: The suggestions for the docstring of `connect_as_extension` go for this one as well.

zospy/zpcore.py
```diff
@@ -282,13 +282,21 @@ class ZOS:
     ----------
         ZOSAPI (None | netModuleObject): The ZOSAPI interface or if loaded.
         ZOSAPI_NetHelper (None | netModuleObject): The ZOSAPI_NetHelper interface if loaded.
+
+    Parameters
```
crnh, line 286: Please put the Parameters section before the Attributes section, in order to adhere to NumPyDoc: https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

zospy/zpcore.py
```diff
@@ -282,13 +282,21 @@ class ZOS:
...
+    ----------
+    preload: bool
```
crnh, line 288: This one applies to all docstrings: NumPyDoc uses a space before the colon.
```suggestion
    preload : bool
```

crnh (changes requested):
Apart from the comments in my review, I noticed the class docstring can be updated with the new connection code.
The class docstring also documents an "Advanced" section which actually calls some private methods. If the user is expected to use these, they should be made public. Otherwise it may be better to remove this section from the docstring (which I prefer). Furthermore, this section mentions the non-existing `ZOS._update_constants` method. @LucVV what do you think?

zospy/zpcore.py
LucVV: I think we cannot directly drop them, as a `TypeError` would be raised [when either of these arguments is specified upon initiating `ZOS`](https://stackoverflow.com/questions/59217884/new-method-giving-error-object-new-takes-exactly-one-argument-the-typ). For example, `zos = ZOS(preload=True)` would raise a `TypeError`. 

This is the result of `*args` and `**kwargs` being passed to `instance = super(ZOS, cls).__new__(cls, *args, **kwargs)`, which results in the arguments of `__init__()` being passed when specified. As the super of ZOS is always `object()`, this raises the `TypeError`.

We could handle this elegantly by for example [checking wheter the base class is `object`](https://github.com/python/typing/pull/517), but I think we have a controlled situation and can simply adjust `instance = super(ZOS, cls).__new__(cls, *args, **kwargs)` to `instance = super(ZOS, cls).__new__(cls)`

I would like to use that option, as it will be able to add additional arguments to `ZOS.__init__()` in the future without it affecting `ZOS.__new__()`.

LucVV:
> Apart from the comments in my review, I noticed the class docstring can be updated with the new connection code. The class docstring also documents an "Advanced" section which actually calls some private methods. If the user is expected to use these, they should be made public. Otherwise it may be better to remove this section from the docstring (which I prefer). Furthermore, this section mentions the non-existing `ZOS._update_constants` method. @LucVV what do you think?

I think this docstring is indeed outdated. I will update it!

zospy/zpcore.py
crnh: Ah, I see I was mistaken. Your preferred option sounds good!

LucVV: I have adjusted this accordingly

zospy/zpcore.py
```diff
@@ -300,7 +300,7 @@ def __new__(cls, *args, **kwargs):
         if len(cls._instances) >= 1:
             # As the number of applications within runtime is limited to 1 by Zemax, it is logical to also limit the
             # number of ZOS instances
-            raise ValueError(
+            raise ConnectionRefusedError(
```
crnh, line 326: I am not sure about this one. Using a `ConnectionRefusedError` suggests that something went wrong when connecting, which simply isn't true. This method doesn't handle anything connection-related, and raising this exception here is therefore misleading. If we really want to change this, raising a `RuntimeError` seems more appropriate.

LucVV: Oke! I will change it to `RuntimeError` as I think `ValueError` was also not appropriate in hindsight.

LucVV: After discussing with @crnh, I reverted the changes of this error type as it might induce some breaking changes, thus, we will change it whenever more breaking changes will be implemented.

andibarg:
I implemented the suggested changes to the docstrings and the function returns. It seems to work in my own scripts but I am not sure why tox complains.

LucVV:
> It seems to work in my own scripts but I am not sure why tox complains.

It failed due to an error in linting, which is now fixed! I will run the unit tests after the new tests are fully implemented.

crnh:
@andibarg Next time, before doing a push, can you please run `tox` locally? The [contribution guidelines](CONTRIBUTING.md) actually require you to do this before opening a PR. By following the contribution guidelines, you help us to focus on things that matter. Thanks!

You can install `tox` using pip or conda:

```shell
pip install tox
```

And then run it in the project directory:

```shell
tox
```

Please note that the GitHub action is not a substitute for running tox locally, because it doesn't fix anything and also can't run the unit tests.

crnh:
@LucVV Testing both initialization methods is indeed complicated. I thought about parameterizing the ZOS fixture, but this will change the names of all tests and therefore mess up our reference data tests.

If it is easy to force some tests to be run first, we can use the tests you proposed (testing the legacy stuff before all other things). However, this probably requires us to introduce a new dependency: [pytest-order](https://pytest-order.readthedocs.io/en/latest/).

If that doesn't work and more complex changes are necessary, this may not be the right moment to introduce these changes. In that case, I think it's better to put this PR on hold and wait until we need to introduce more fundamental changes.

LucVV:
After a discussion with @crnh, we decided that one way to move forward is to implement an option for pytest (`--legacy-connection-setup`) that specifies whether you want to use either the old method to connect (legacy), or this method. I have implemented this option.

Furthermore, I have added one standalone environment to `tox.ini` that specifically runs the `test_zpcore` with `--legacy-connection-setup`. If these tests pass, the other tests should behave exactly as they would without `--legacy-connection-setup`.

crnh:
I just fixed some minor issues with the docstrings and ran the tests, but some of them fail in extension mode. In standalone mode, all tests pass. However, this looks like it is caused by a bug in the ZOS-API, so I think this one is good to go.

The problem is caused by calling `Application.GetSystemAt(pos)` in `ZOS.get_system`. Even when called with `pos = 0`, this method fails with an `System.ArgumentOutOfRangeException`. I looked further into this and `Application.NumberOfOpticalSystems` turns out to be 0 when connecting in extension mode, which seems weird since the primary system is always present. I do not think this is related to ZOSPy, and will report this on the Zemax forum.

crnh (approved):

LucVV (approved):

PR #27, opened 2023-07-17
Scientific notation regular expression for get_number_field in polarization

andibarg:
I got a wrong result from the polarization transmission function. For small values the output text file from OpticStudio includes numbers in scientific notation and ` _get_number_field()` currently does not catch that. I updated the regular expression accordingly.


LucVV:
Thank you for reporting and fixing this! Do you perhaps have a case where this happens (e.g. by using specific settings)? That way, we can also add it to a unit test and assure that all works as intended.

andibarg:
@LucVV The simplest case would be (nearly) cross polarization. We just need a single Jones matrix surface set as a polarizer, so Areal = 1 and 0 for the rest. The Jones vector in the transmission could be something like Jx = 1e-5 and Jy=1. Then the transmission is so small that Zemax switches to scientific notation.

@crnh At first, I was quite sure that the exponent is always two digits, in which case both `\d?\d?` and `\d{0,2}` work equally well. However, after checking this again I find that three digits are also possible but never more. Good that you pointed it out, I will change the regex to `\d{0,3}` now. Though we should make sure that the output is not just specific to my PC again.

crnh:
Great that we can do this using a Jones matrix! Our current tests for the polarization analyses are using a Jones matrix as well, so this allows us to incorporate this scenario in our existing tests.

Regarding the regex, it would currently match e.g. `1.23E`, since it allows the exponent to have zero digits. `\d{1,3}` might be better 😄. Furthermore, I would vote for the slightly more generic `\d+`. I don't think allowing an unlimited amount of digits is going to cause problems, and this also covers very obscure cases where the number of digits in the exponent might turn out to not be limited to 3.

zospy/analyses/polarization.py
```diff
@@ -19,7 +19,7 @@
 
 
 def _get_number_field(name: str, text: str) -> str:
-    return re.search(rf"{name}\s*:\s*(-?[\d{_config.DECIMAL_POINT}]+)", text).group(1)
+    return re.search(rf"{name}\s*:\s*(-?[\d{_config.DECIMAL_POINT}]+[Ee]?[+-]?\d{0,3})", text).group(1)
```
LucVV, line 22: For me, the this does not work, to make it work, the last part has to be updated to `\d{{0,3}}` to escape the curly brackets in the formatted raw string. 

However, as @crnh also indicated, I think the regex for the exponent can be updated. In addition, I also think we should maybe use `re.escape()` for both `name` and `_config.DECIMAL_POINT` to prevent including accidental regex patterns. Thus, I would like to change the regex pattern to:

```suggestion
    return re.search(rf"{re.escape(name)}\s*:\s*([-+]?\d+{re.escape(_config.DECIMAL_POINT)}?\d*(?:[Ee][-+]?\d+)?)", text).group(1)

```

LucVV:
With regards to unit testing, I found out that running `test_transmission_returns_correct_result` with the following settings `("32x32", False, 0.001, 1, 0, 0)` would result in `Data.TotalTransmission` being an exponential number. As the test evaluates `Data.TotalTransmission`, this would add a test for the exponential numbers with minimal change.

LucVV:
I have added the unit test, @crnh do you want to add a separate test for `_get_number_field()`, or do you think this suffices?

crnh:
I would indeed like to have separate tests. The idea of unit tests is to test small pieces of code ("units") that perform a single task. Testing this separately makes it easier to distinguish between problems with parsing and problems with running the analysis in general. A separate test can probably also be re-used when we start working on more generic analysis parsers.

tests/analyses/test_polarization.py
```diff
@@ -1,8 +1,49 @@
...
+
+_signs = ["", "+", "-"]
+_int_numbers = ["1", "123"]
+_float_numbers = [".1", ".123", "1.", "1.2", "1.23", "12.3"]
+_decimal_separators = [",", "."]
+_exps = ["", "e1", "e123", "e+1", "e+123", "e-1", "e-123", "E1", "E123", "E+1", "E+123", "E-1", "E-123"]
+
+
+class TestGetNumberField:
+    @pytest.mark.parametrize(
```
crnh, line 19: It is possible to add multiple paramterize decorators to a single test, which effectively generates a test matrix and looks cleaner.

```suggestion
    @pytest.mark.parametrize("sign", _signs)
    @pytest.mark.parametrize("number", _int_numbers)
    @pytest.mark.parametrize("exp", _exps)
    def test_get_number_field_returns_correct_result_for_integers(self, sign, number, exp):
        ...
```

tests/analyses/test_polarization.py
```diff
@@ -1,8 +1,49 @@
...
+            ((_sign + _number + _exp).replace(".", _decimal_separator), _decimal_separator)
+            for _sign in _signs
+            for _number in _float_numbers
+            for _exp in _exps
+            for _decimal_separator in _decimal_separators
+        ],
+    )
+    def test_get_number_field_returns_correct_result_for_floats(self, number_string, decimal_separator):
+        original_decimal_point = _config.DECIMAL_POINT  # store original decimal separator to revert later
+        if decimal_separator is not None:
```
crnh, line 39: `decimal_separator` can't be `None`

tests/analyses/test_polarization.py
```diff
@@ -1,8 +1,49 @@
...
+            _config.DECIMAL_POINT = decimal_separator  # adjust configured decimal point separator
```
crnh, line 40: Mock testing using `pytest.monkeypatch` can be used to temporarily change parts of a module: https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

```suggestion
        monkeypatch.setattr(_config, "DECIMAL_POINT", decimal_separator)
```

tests/analyses/test_polarization.py
```diff
@@ -1,8 +1,49 @@
...
+
+        res = _get_number_field("Test", f"Test: {number_string}")
+
+        _config.DECIMAL_POINT = original_decimal_point  # restore originally configured decimal point
```
crnh, line 44: Not necessary if you use monkeypatch

crnh (changes requested):
Regarding the test names: `TestGetNumberField` is already present in the test class name, so I think `test_parses_int` and `test_parses_float` are descriptive enough

crnh:
Looks good! All tests pass on my system. The general comment from my previous review has not been addressed, but that is not urgent / mandatory.

The current unit tests do not cover differences in whitespace around the colon, but I don't think it is important to add tests for these now.

crnh (dismissed):

LucVV:
I changed the name of the functiosn as request but forgot to commit & push before the review. I have dismissed the review in a hope it would show again in this PR but it seems to be missing still.

crnh (approved):

LucVV:
Hi @andibarg, the PR has been merged! Thank you again for reporting this bug and helping us to solve it!

LucVV:
@crnh I have missed that we did not add this pull request to the changelog. Should we create a new pull request to do so or do we keep track of it and add it when we create a new release?

crnh:
I think you can edit the file directly on the master branch, which is fine in this case. If not, you can create a new PR.

PR #40, opened 2023-09-25
Fix LensUpdateMode

LucVV:
## Proposed change

The setter for `LensUpdateMode` does not work currently, as it is trying to set `OpticStudioSystem.LensUpdateMode` rather than 'OpticStudioSystem.UpdateMode`. There is some ambiguity as the contact is called 'LensUpdateMode` but the ZOS-API calls it `UpdateMode`. For now, I think we should at least fix 'LensUpdateMode'. 

We can also choose to add 'UpdateMode' and drop LensUpdateMode), but that would be a breaking change. If we would want that, I would propose to fix LensUpdateMode for now, add `UpdateMode` and add a deprecation warning to `LensUpdateMode`. I would like some input on that.

## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ ZOS-API can also be added in patch releases.
 
 ### Fixed
 
-- Bug that did not allow users to change the LensUpdateMode directly through `OpticStudioSystem.LensUpdateMode`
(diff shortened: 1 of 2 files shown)


crnh (approved):
In the future, we may want to change this attribute to `UpdateMode`, to be more in line with the ZOS-API names.

PR #55, opened 2023-12-10
Huygensmtf

noahrbn:
## Type of change

- [x] New analysis (a wrapper around an OpticStudio analysis)

# Additional information

- Python version: ...
- OpticStudio version: ...

## Checklist

### DIFF
--- a/zospy/analyses/mtf.py
+++ b/zospy/analyses/mtf.py
@@ -237,3 +237,118 @@ def fft_through_focus_mtf_fromcfg(
     )
 
     return analysis.complete(oncomplete, result)
+
+def huygens_mtf(
+    oss: OpticStudioSystem,
+    pupilsampling: str | int = "32x32",
+    imagesampling: str | int = "32x32",
+    imagedelta: float = 0.0,
+    wavelength: str | int = "All",
+    field: str | int = "All",
+    mtftype: constants.Analysis.Settings.Mtf.HuygensMtfTypes | str = "Modulation",
+    maximumfrequency: float = 150.0,
+    use_polarization: bool = False,
+    use_dashes: bool = False,
+    oncomplete: OnComplete | str = OnComplete.Close,
+) -> AnalysisResult:
+    """Wrapper around the OpticStudio Huygens MTF.
+
+    For an in depth explanation of the parameters, see the Zemax OpticStudio user manual
+
+    Parameters
+    ----------
+    oss: zospy.core.OpticStudioSystem
+        A ZOSPy OpticStudioSystem instance. Should be sequential.
+    pupilsampling: str | int
+        The pupil sampling, either string (e.g. '64x64') or int. The integer will be treated as a ZOSAPI Constants integer.
+    imagesampling: str | int
+        The image sampling, either string (e.g., '64x64') or int. The integer will be treated as a ZOSAPI Constants integer.
+    imagedelta: float
+        The Image Delta, defaults to 0.0.
+    wavelength: str | int
+        The wavelength to use in the MTF. Either 'All' or an integer specifying the wavelength number.
+    field: str | int
+        The field to use in the MTF. Either 'All' or an integer specifying the field number.
+    mtftype: zospy.constants.Analysis.Settings.Mtf.MtfTypes.Modulation
+        The MTF type (e.g. `Modulation`) that is calculated.
+    maximumfrequency: float
+        The maximum frequency at which the MTF is calculated in lp/mm. Defaults to 150.0.
+    use_polarization: bool
+        Use polarization. Defaults to False.
+    use_dashes: bool
+        Use dashes. Defaults to False.
+    oncomplete: OnComplete | str
+        Defines behaviour upon completion of the analysis. Should be one of ['Close', 'Release', 'Sustain']. If 'Close',
+        the analysis will be closed after completion. If 'Release', the analysis will remain open in OpticStudio, but
+        the link with python will be destroyed. If 'Sustain' the analysis will be kept open in OpticStudio and the link
+        with python will be sustained. To enable interaction when oncomplete == 'Sustain', the OpticStudio Analysis
+        instance will be available in the returned AnalysisResult through AnalysisResult.Analysis. Defaults to 'Close'.
+
+    Returns
+    -------
+    AnalysisResult
+        A HuygensMTF analysis result
+    """
+    analysis_type = constants.Analysis.AnalysisIDM.HuygensMtf
+
+    # Create analysis
+    analysis = new_analysis(oss, analysis_type)
+
+    # Apply settings
+    analysis.Settings.PupilSampleSize = getattr(constants.Analysis.SampleSizes, utils.zputils.standardize_sampling(pupilsampling))
+    analysis.Settings.ImageSampleSize = getattr(constants.Analysis.SampleSizes, utils.zputils.standardize_sampling(imagesampling))
+    analysis.Settings.ImageDelta = imagedelta
+    analysis.set_wavelength(wavelength)
+    analysis.set_field(field)
+    analysis.Settings.Type = constants.process_constant(constants.Analysis.Settings.Mtf.HuygensMtfTypes, mtftype)
+    analysis.Settings.UsePolarization = use_polarization
+    analysis.Settings.UseDashes = use_dashes
+    analysis.Settings.MaximumFrequency = maximumfrequency
+
+    # Correct an API bug in setting API type for OpticStudio version <21.2
+    # TODO: not sure if this is needed here; commented for now
+    #_correct_fft_through_focus_mtftype_api_bug(oss, analysis, mtftype)
+
+    # Calculate
+    analysis.ApplyAndWaitForCompletion()
+
+    # Get headerdata, metadata and messages
+    headerdata = analysis.get_header_data()
+    metadata = analysis.get_metadata()
+    messages = analysis.get_messages()
+
+    # Get settings
+    settings = pd.Series(name="Settings", dtype=object)
+
+    settings.loc["PupilSampling"] = str(analysis.Settings.PupilSampleSize)
+    settings.loc["ImageSampling"] = str(analysis.Settings.ImageSampleSize)
+    settings.loc["ImageDelta"] = analysis.Settings.ImageDelta
+    settings.loc["Wavelength"] = analysis.get_wavelength()
+    settings.loc["Field"] = analysis.get_field()
+    settings.loc["Type"] = str(analysis.Settings.Type)
+    settings.loc["UsePolarization"] = analysis.Settings.UsePolarization
+    settings.loc["UseDashes"] = analysis.Settings.UseDashes
+    settings.loc["MaximumFrequency"] = analysis.Settings.MaximumFrequency
+
+    # Get data and unpack
+    data = []
+    for ii in range(analysis.Results.NumberOfDataSeries):
+        data.append(utils.zputils.unpack_dataseries(analysis.Results.DataSeries[ii]))
+
+    if not len(data):
+        data = pd.DataFrame()
+    elif len(data) == 1:
+        data = data[0]
+    else:
+        data = pd.concat(data, axis=1)
+
+    result = AnalysisResult(
+        analysistype=str(analysis_type),
+        data=data,
+        settings=settings,
+        metadata=metadata,
+        headerdata=headerdata,
+        messages=messages,
+    )
+
+    return analysis.complete(oncomplete, result)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 with one exception: small features that only simplify access to certain parts of the
 ZOS-API can also be added in patch releases.
 
+## [Unreleased]
+
+### Added
+- MTF analysis: `huygens_mtf` (#50)
+
 ## [[1.1.1]](https://github.com/MREYE-LUMC/ZOSPy/releases/tag/v1.1.1) - 2023-09-25
 
 ### Added


crnh:
Hi @noahrbn,

Thank you for contributing to ZOSPy! As it's quite busy on our side, it may take some time before we can review your contribution. Meanwhile, can you lint, format and test your code [as indicated in the contribution guidelines](https://github.com/MREYE-LUMC/ZOSPy/blob/main/CONTRIBUTING.md#workflow)? The output of the pull request checks suggests there are some issues related to our coding conventions that can be easily resolved by following the workflow explained in the contribution guidelines. If you are new to this and don't know how to get started, feel free to drop us a message and we're happy to help you!

noahrbn:
The lint checks on my end are now passing - I will write unit tests (still familiarizing myself with your unit testing conventions) in another pull request. Please feel free to say anything obvious - I am somewhat unfamiliar with this.

I see when I run `tox` black lint automatically reformats the code to pass, but ruff lint does not. Is there a flag to pass to `tox` to automatically reformat to fix any errors?

crnh:
Nice to hear you got `tox` running! Black does indeed automatically reformat the code. For `ruff`, you can pass the `--fix` argument through `tox` (note the extra double dashes `--` that separate the arguments for `ruff` from the arguments for `tox`):

```shell
tox run -e lint -- --fix
```

Note that not all errors can be fixed automatically (errors that can, are marked with `[*]`). Looking at the output from the GitHub Actions, most of them should already be solved by running `black`.

As for the unit tests, we'll wait for your 2nd PR. As the `huygens_mtf` analysis is largely similar to `fft_through_focus_mtf`, the unit tests will probably similar as well. Please also take a look at the [documentation for generating reference data](https://github.com/MREYE-LUMC/ZOSPy/blob/main/tests/README.md#generating-test-reference-data), as you need some data to test against ;).

noahrbn:
I decided to include the unit testing as a part of this PR. I have committed it now and the new tests under `test_mtf.py` seem to pass on my machine. Let me know if anything else is needed to approve the PR.

I have a few questions:

1. Could you clarify exactly how this unit testing is set up? I don't understand it from your README documents. Basically, what I understand is that the `tests.yaml` file specifies all the tests that are to be performed, and the `generate_test_data_reference` script goes through this file and creats `.zmx` files specific to each test (for most of them it seems to be `simple_system`, i.e., just a singlet lens). The analysis is run on this file. Then, when `pytest` is run, we see if the analysis performed with `zospy` matches the results from the file. Maybe I'm missing something, but isn't this a bit circular? What exactly is being tested here?
2.  How does the communication between `pytest`/each `test_*.py` file and the files in `tests/data/reference` work? For example, in each `test_*_returns_correct_results` method, it is unclear to me how `expected_data` is passed in.
3. Why is it necessary to have both a `test_*_returns_correct_results` and `test_*_matches_reference_data` method? If I look at how `fft_through_focus_mtf` was implemented in `tests.yaml`, it seems that only the `correct_results` method is tested against reference data.

crnh:
Thanks a lot for contributing the unit tests as well, nice work!

Anwers to your questions:

1. The `generate_reference_data` script is separated from the unit tests. It is only used to generate the data files that are used by the unit tests to check the output of the analyses. `tests.yaml` is only used by the `generate_reference_data` script. It specifies for each analysis on which optical system it should be run, and which arguments should be supplied to the analysis (often there are multiple sets of parameters for a single analysis). For each specified analysis and parameter set, `generate_reference_data` builds the system in OpticStudio, performs the analysis and saves the analysis output using `zospy.analyses.base.AnalysisResult.to_json()`. It also saves the system in `.zmx` format. Only the json files are used; the `.zmx` files are only saved for reference purposes.

    Isn't this a bit circular? Yes, it is: we use ZOSPy to generate the reference analysis output, and then test ZOSPy against this output. This is not really a "unit test", unlike for example `test_to_json`, which clearly tests a single piece of functionality (the ability to reconstruct an `AnalysisResult` from its own json serialization). Isn't that useless? No, it isn't: these tests should be seen as regression tests. Although we would like to, we cannot really test the validity of OpticStudio's output, but at least we can check if it stays the same. If we change some functionality used by one or more analyses, these tests help us to make sure the change didn't break the analyses or change their output.

2. To pass the expected data to the unit tests, we make use of PyTest's [fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html). These fixtures are functions that can perform some setup steps and then return a value; this helps to keep the unit tests modular. As you can see, `tests/analyses/testconf.py` defines a fixture named [`expected_data`](https://github.com/MREYE-LUMC/ZOSPy/blob/e647c2b12c87e0af8b0acfa4f3f40025c05f7c0c/tests/analyses/conftest.py#L17C5-L17C18). When PyTest encounters the `expected_data` parameter in the test definition, it will call the `expected_data` fixture and pass its returned value as an argument to the unit test. If you check this fixture, you will see that it determines the name of the appropriate reference data file based on the test name and its parameters. If this file is not present, the test will be skipped.

    Note that this also means you _need_  to run the `generate_reference_data` script when you add new tests; otherwise these tests will still be skipped, because the necessary reference data is missing.

3. We do not only want to ensure consistent results for a single version of OpticStudio, but also check if results are consistent between different versions. To do this, we defined a reference version (currently OpticStudio 23 R1). If you run the unit tests with a different OpticStudio version (say, 23 R2), `test_*_returns_correct_results` will check the analysis output against the reference data for that version (OpticStudio 23 R2). `test_*_matches_reference_data` will check the analysis output against the reference data for the reference version (OpticStudio 23 R1). For this check, the reference data files for OpticStudio 23 R1 are used, so it is not necessary to separately specify these test cases for the `test_*_matches_reference_data` tests (you can check this if you take a look at the [`reference_data`](https://github.com/MREYE-LUMC/ZOSPy/blob/e647c2b12c87e0af8b0acfa4f3f40025c05f7c0c/tests/analyses/conftest.py#L30) fixture).

I hope this answers your questions, and let me know if something is still unclear!

noahrbn:
Thanks for the clarification - that makes a lot of sense.

Please let me know if any more information is needed to approve the PR.

LucVV:
I see that `test_huygens_mtf_returns_correct_result` uses `assert np.allclose` to compare the result and the expected data. We have just released version 1.1.2 and one of the updates is that we changed to `pandas.testing.assert_frame_equal` for these comparisons as that also checks the columns and indices of the DataFrames (see #56). I think that should also be the way forward in this merge request.

tests/analyses/test_mtf.py
```diff
@@ -86,3 +90,33 @@ def test_fft_through_focus_mtf_sets_mtftype_correctly(self, simple_system, mtfty
...
+        result = huygens_mtf(
+            simple_system,
+            pupilsampling=pupilsampling,
+            imagesampling=imagesampling,
+            imagedelta=imagedelta,
+            mtftype=mtftype,
+            maximumfrequency=maximumfrequency,
+        )
+
+        assert np.allclose(result.Data.astype(float), expected_data.Data.astype(float), rtol=1e-3)
```
LucVV, line 122:
```suggestion
       assert_frame_equal(result.Data.astype(float), expected_data.Data.astype(float))
```

tests/analyses/test_mtf.py
```diff
@@ -1,7 +1,11 @@
 import numpy as np
 import pytest
```
LucVV, line 3:
```suggestion
from pandas.testing import assert_frame_equal

```

noahrbn:
@LucVV I have accepted and committed your suggestions. I will test this branch locally myself to make sure the tests still pass, but other than that I think this is good to go.

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
     )
 
     return analysis.complete(oncomplete, result)
+
+
+def huygens_mtf(
+    oss: OpticStudioSystem,
+    pupilsampling: str | int = "32x32",
```
crnh, line 244:
```suggestion
    pupil_sampling: str | int = "32x32",
```

This keeps the naming consistent with other analyses (check e.g. [`psf.huygens_psf`](https://github.com/MREYE-LUMC/ZOSPy/blob/8b0d401870c36541df7117ebb18cdb764026740b/zospy/analyses/psf.py#L14)).
Please also update all other occurrences (e.g. in the unit tests and test reference data specification).

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
...
+    imagesampling: str | int = "32x32",
```
crnh, line 245:
```suggestion
    image_sampling: str | int = "32x32",
```

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
...
+    imagedelta: float = 0.0,
+    wavelength: str | int = "All",
+    field: str | int = "All",
+    mtftype: constants.Analysis.Settings.Mtf.HuygensMtfTypes | str = "Modulation",
+    maximumfrequency: float = 150.0,
```
crnh, line 250:
```suggestion
    maximum_frequency: float = 150.0,
```

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
     )
 
     return analysis.complete(oncomplete, result)
+
+
+def huygens_mtf(
+    oss: OpticStudioSystem,
+    pupilsampling: str | int = "32x32",
+    imagesampling: str | int = "32x32",
+    imagedelta: float = 0.0,
```
crnh, line 246:
```suggestion
    image_delta: float = 0.0,
```

See previous comment; this name is also used in [psf.huygens_psf](https://github.com/MREYE-LUMC/ZOSPy/blob/8b0d401870c36541df7117ebb18cdb764026740b/zospy/analyses/psf.py#L16).

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
...
+    use_dashes: bool = False,
+    oncomplete: OnComplete | str = OnComplete.Close,
+) -> AnalysisResult:
+    """Wrapper around the OpticStudio Huygens MTF.
+
+    For an in depth explanation of the parameters, see the Zemax OpticStudio user manual
+
+    Parameters
+    ----------
+    oss: zospy.core.OpticStudioSystem
```
crnh, line 261: NumpyDoc requires a space before the colon:

```
    oss : zospy.core.OpticStudioSystem
```

Unfortunately, I have not yet found a way to check this automatically (our formatting tools can't do it, as far as I know), so we still need to change this manually.

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
...
+    analysis.Settings.ImageDelta = imagedelta
+    analysis.set_wavelength(wavelength)
+    analysis.set_field(field)
+    analysis.Settings.Type = constants.process_constant(constants.Analysis.Settings.Mtf.HuygensMtfTypes, mtftype)
+    analysis.Settings.UsePolarization = use_polarization
+    analysis.Settings.UseDashes = use_dashes
+    analysis.Settings.MaximumFrequency = maximumfrequency
+
+    # Correct an API bug in setting API type for OpticStudio version <21.2
+    # TODO: not sure if this is needed here; commented for now
```
crnh, line 316: This needs to be checked before we can merge this PR. @LucVV you are still running a version that's old enough, can you check this? Maybe we first need to extend the unit tests, because they do not yet change this parameter.
And @noahrbn which OpticStudio version do you use?

tests/analyses/test_mtf.py
```diff
@@ -86,3 +91,33 @@ def test_fft_through_focus_mtf_sets_mtftype_correctly(self, simple_system, mtfty
         )
 
         assert result.Settings["Type"] == _FFT_THROUGH_FOCUS_MTF_MTFTYPE_EXPECTED_RETURN[mtftype]
+
+
+class TestHuygensMTF:
+    def test_can_run_fft_through_focus_mtf(self, simple_system):
```
crnh, line 97:
```suggestion
    def test_can_run_huygens_mtf(self, simple_system):
```

tests/analyses/test_mtf.py
```diff
@@ -86,3 +91,33 @@ def test_fft_through_focus_mtf_sets_mtftype_correctly(self, simple_system, mtfty
...
+        assert result.Data is not None
+
+    def test_to_json(self, simple_system):
+        result = huygens_mtf(simple_system)
+
+        assert result.from_json(result.to_json())
+
+    @pytest.mark.parametrize(
+        "pupilsampling,imagesampling,imagedelta,mtftype,maximumfrequency",
+        [("64x64", "64x64", 0.0, "Modulation", 150.0)],
```
crnh, line 109: Please add multiple sets of parameters; if possible, every parameter should be changed at least once. In the past, this has helped us to e.g. detect the problem with changing the MTF type for the FFT through focus MTF analysis.

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,125 @@ def fft_through_focus_mtf_fromcfg(
...
+    imagedelta: float
+        The Image Delta, defaults to 0.0.
+    wavelength: str | int
+        The wavelength to use in the MTF. Either 'All' or an integer specifying the wavelength number.
+    field: str | int
+        The field to use in the MTF. Either 'All' or an integer specifying the field number.
+    mtftype: zospy.constants.Analysis.Settings.Mtf.MtfTypes.Modulation
+        The MTF type (e.g. `Modulation`) that is calculated.
+    maximumfrequency: float
+        The maximum frequency at which the MTF is calculated in lp/mm. Defaults to 150.0.
```
crnh, line 278: > lp/mm

Maybe a stupid question, but what is this unit? (I don't use this type of analysis in my own research).

crnh: I also see OpticStudio allows to change this unit, so I don't think it should be in the docstring.

crnh (changes requested):
Very nice work, thanks a lot! There are some minor issues that need to be addressed before we can merge this PR.

zospy/analyses/mtf.py
LucVV: The [post](https://community.zemax.com/zos-api-12/zos-api-setting-mtf-property-type-not-working-730) about this on the opticstudio forum only mentions `fft_through_focus_mtf`. However, I estimate that the fix might be required here also. I'll look into this

LucVV: I now see that `HuygensMtf` uses the enum `HuygensMtfTypes` rather than `MtfType`. In my version of OpticStudio, it is an enum that only contains `Modulation`. Thus, there is no way (and no need) to test this bugfix. However, I am in favor of keeping `mtftype` as parameter for future us.

crnh: I checked this in OpticStudio 23 R2 and there the enum also has a single value. This means these lines can be removed.
I also prefer `mtftype` as parameter name over (variants of) `huygens_mtftype`.

zospy/analyses/mtf.py
noahrbn: Is this extra space required in the entire docstring? Existing docstrings in the same file do not have this.

zospy/analyses/mtf.py
noahrbn: This stands for line pairs per mm. I don't see where OpticStudio allows one to change this unit for the Huygens MTF analysis.

crnh: The OpticStudio documentation states the frequency is in "MTF Units". These can be adjusted in the System Explorer under Units.

All available units are Cycles / something; are line pairs the same as cycles?

noahrbn: I believe they are, but we can label it cycles. You are right - you can change between cycles/mm and cycles/milliradian. I think the method would still work if the system had been configured for cycles/milliradian, but the `maximum_frequency` argument default value should perhaps be a function of the "MTF Units" set in the system - I am not sure if this is possible in Python, actually.

Maybe this is something for another PR - units of cycles/(angular quantity) are somewhat uncommon from what I can tell.

zospy/analyses/mtf.py
noahrbn: I have made the requested variable name changes and propagated them to the unit tests and test reference data specification.

zospy/analyses/mtf.py
noahrbn: I have modified the docstring of `huygens_mtf` accordingly.

tests/analyses/test_mtf.py
noahrbn: I have added three additional unit tests.

noahrbn:
I have addressed the above comments with the most recent commits. I have linted the code and all unit tests (including new ones) pass on my machine.

zospy/analyses/mtf.py
noahrbn: For now, I have addressed this by modifying the docstring in https://github.com/MREYE-LUMC/ZOSPy/pull/55/commits/ef9f57c2d82a58a463c9fbe66711de0fa51a9f6c.

crnh: I think it's best to refer to "MTF units" and mention the available units, along the line of what you did (and then hopefully Ansys won't add more units, because we'll probably forget to add them to this docstring...)

In my opinion, the MTF unit should not be modified from an analysis, because this has system-wide effects. The user should do this explicitly from `oss.SystemData`.

tests/analyses/test_mtf.py
crnh: This has not yet been addressed

tests/analyses/test_mtf.py
crnh: I see you added the specification for the reference data as well, but that commit seems somehow to be lost. At least I can see only a single set of parameters when I open this PR in VS Code for the web. Can you look into this?

zospy/analyses/mtf.py
```diff
@@ -237,3 +237,123 @@ def fft_through_focus_mtf_fromcfg(
...
+        The Image Delta, defaults to 0.0.
+    wavelength : str | int
+        The wavelength to use in the MTF. Either 'All' or an integer specifying the wavelength number.
+    field : str | int
+        The field to use in the MTF. Either 'All' or an integer specifying the field number.
+    mtftype : zospy.constants.Analysis.Settings.Mtf.MtfTypes.Modulation
+        The MTF type (e.g. `Modulation`) that is calculated.
+    maximum_frequency : float
+        The maximum frequency at which the MTF is calculated.
+        Units are either cycles/mm or cycles/mr, depending on system setting).
```
crnh, line 279: Unmatched closing parenthesis:

> ```setting).```

crnh (changes requested):
Thanks for addressing my comments! I think this one is ready to be merged if you address these final comments and all tests pass on our systems.

zospy/analyses/mtf.py
noahrbn: Fixed in https://github.com/MREYE-LUMC/ZOSPy/pull/55/commits/d566fe3322ca9ca320e498f5c30960894103b553.

tests/analyses/test_mtf.py
noahrbn: Sorry, somehow I forgot to commit that change. This is addressed in https://github.com/MREYE-LUMC/ZOSPy/pull/55/commits/d566fe3322ca9ca320e498f5c30960894103b553. All unit tests pass on my machine.

crnh (approved):
All tests pass on my system! Thanks for and congratulations with your first contribution, @noahrbn! If you want, you may add the reference data files for your OpticStudio version to this PR, but we do not require this.

@LucVV Do you still want to run the unit tests on your system?

LucVV:
Great to hear! And congratulations indeed @noahrbn!

I'll run the test and add reference data for OpticStudio 20.3.2 today!

LucVV:
The new tests run on my system and the test results of OpticStudio 20.3.2 match those of 23.2.1! I have added the reference data for 20.3.2

I noticed that only `test_huygens_mtf_returns_correct_result` was implemented. I have also implemented `test_huygens_mtf_matches_reference_data` to compare different OpticStudio versions. @crnh, can you add the reference files for 23.1.0 (the current reference version)?

LucVV (approved):
All looks good! @crnh should we still add reference data for 23.1.0?

crnh:
Yes, I'll add the reference data today!

PR #56, opened 2023-12-12
Fix reversed row index in datagrids

### LINKED ISSUE #54: PSF data grid does not seem to be centered correctly
### Problem

While using the Huygens PSF analysis for a rotationally symmetric system, I noticed that the maximum value of the PSF is not at (0,0). This is in contrast to the output directly taken from OpticStudio. I provided a simplified example below to reproduce this. And in case you are wondering, it seems to have nothing to do with the "use centroid" option.

As far as I can see, the problem originates from the `unpack_datagrid` function in `zputils.py`.  In row 104, we have:
```python
df = pd.DataFrame(data=values, index=rows[::-1], columns=columns)
``` 
I inverted the array `rows`, which gave me the correct results:
```python
df = pd.DataFrame(data=values, index=rows, columns=columns)
``` 
What is the reason for inverting `rows` in the first place? Do you see any issues with my solution? Is `unpack_datagrid` used by any other analysis that should be checked?

### What version of ZOSPy are you running?

1.1.1

### What version of OpticStudio are you running?

23.2.1

### Which operating system do you use?

Windows 10

### In which environment do you use ZOSPy?

- [X] Plain Pythons scripts (not in Spyder)
- [ ] Jupyter notebooks
- [ ] Spyder

### In which connection modes does the problem occur?

- [ ] Extension mode
- [X] Standalone mode

### Example code

```Python
import os
import numpy as np
import zospy as zp
import matplotlib.pyplot as plt

# Load Zemax file
zos = zp.ZOS()
oss = zos.create_new_application(True)
oss.load(os.path.join(zos.Application.SamplesDir,
                      'Sequential',
                      'Objectives',
                      'Double Gauss 28 degree field.zmx')

# Get Huygens PSF
psfdata = zp.analyses.psf.huygens_psf(oss).Data

# Get center and max value
print(psfdata[0][0])
print(np.max(psfdata))

# Plot
plt.pcolormesh(psfdata.keys(),psfdata[0].keys(),psfdata)
plt.show()
```

### Output

_No response_

### Additional information

_No response_

crnh:
## Proposed change

The row index of datagrids is currently reversed, resulting in incorrect values being returned when accessing values using their indices (e.g. using `pandas.DataFrame.loc`).

## Type of change

# Additional information

- Python version: 3.9-3.11
- OpticStudio version: 23 R2.1

## Related issues

Closes #54

## Checklist


jwmbeenakker:
Thank you @LucVV and @crnh for fixing this bug. All looks good.

jwmbeenakker (approved):

PR #78, opened 2024-10-11
Implement new analysis structure with parsers

crnh:
## Proposed change

This is a proposal for a new analysis structure, which should make it easier to create new analyses. The new structure and parsers have been implemented for 4 analyses: Surface data, Zernike standard coefficients, Polarization transmission and Single ray trace.
The analyses have been added in `zospy.analyses.new.parsers`, a temporary submodule that allows to test these analyses without introducing breaking changes.

### Parsers

Parsers for text output are based on [Lark](https://lark-parser.readthedocs.io/en/stable/). 
Lark grammars are stored in `zospy/analyses/new/parsers/grammars/`.
This folder contains a file with grammar definitions (`zospy.lark`) for structures commonly found in OpticStudio's text output, and a more specific grammar for each analysis.

Producing an analysis result is a three-stage process:

1. The text output is parsed based on the grammar, which creates a parse tree;
2. The parse tree is processed using a [transformer](https://lark-parser.readthedocs.io/en/stable/visitors.html#transformer), which converts the tree into a dictionary. This dictionary retains the names used in the text output as much as possible;
3. The dictionary is converted to an object using [Pydantic dataclasses](https://docs.pydantic.dev/latest/concepts/dataclasses/).

### Analysis structure

Over a year ago it was proposed to make a decorator that takes care of all the common processing required for all analyses (#15). 
This seemed like a good idea, until I started implementing it.
Writing a decorator that fully takes care of all code currently duplicated across multiple analyses would require the decorator to change the signature of the analysis, which would require some extra work to get correct type hints, and create an inconsistency between the analyses' signatures and docstrings.
All in all, this really looks like we were trying to avoid an object-oriented interface where an object-oriented interface would be the best solution.
I therefore propose a new, object-oriented interface to analyses which looks like this:

```python
import zospy as zp

...

# In one line
result = zp.analyses.raysandspots.SingleRayTrace().run(oss)

# Keep the analysis object
analysis = zp.analyses.raysandspots.SingleRayTrace()
result = analysis.run(oss)
```

All common logic (processing of config and output files, initializing the analysis, creating the analysis result object, finalizing the analysis) is now done in the base class `zospy.analyses.new.base.AnalysisWrapper` (may be renamed to `BaseAnalysisWrapper` to make clear that it is an abstract base class). Each analysis is a child class that only needs to implement the `run_analysis` method, which contains the analysis-specific code, and the `settings` object.

**Other changes:**

- Analysis settings are now separate objects, using Pydantic to validate their attributes (we are using Pydantic anyways for the new analysis results). I chose to do this because this makes it easier to add the settings to the analysis result.
- When `oncomplete == "Sustain"`, the OpticStudio analysis is no longer added to the analysis result. The new analysis object already contains a reference to the OpticStudio analysis.
- Every analysis now has its own file, i.e. the file structure becomes like this:
   ```
   │   base.py
   │   __init__.py
   │
   ├───polarization
   │       transmission.py
   │       __init__.py
   │
   ├───raysandspots
   │       single_ray_trace.py
   │       __init__.py
   │
   ├───reports
   │       surface_data.py
   │       __init__.py
   │
   └───wavefront
         zernike_standard_coefficients.py
         __init__.py
   ```

**Design considerations:**

- [x] I chose to still add the analysis settings to the result object. I think this makes sense, because it allows to easily keep track of the settings that were used to get the result. However, if we remove this dependency, it is no longer necessary to have a separate settings object.
- [x] In this proposal, the constructor for each analysis allows to specify the settings. However, this requires to duplicate the default values. For analyses with many parameters (like the ones @LucVV is working on), this results in rather unwieldy constructors. Alternatively, we can
   1. Remove these parameters from `__init__`;
   2. Instead allow to specify `**kwargs` in `__init__`, which are then passed to the settings object. This way, Intellisense etc. won't show suggestions for parameters anymore, but it is better than option 1.
   3. Choose option 2 only if the number of parameters is large, but this creates a weird asymmetry between analyses.
- [x] How do we want to pass parameters like the config file, text output file, and oncomplete? We can do this in the constructor, together with the settings, but I think it's better to do this in the `run` method (so its signature becomes `run(oss, config_file, text_output_file, oncomplete)`, because these parameters are only used in the `run` method. I would like to know your opinions on this.

## Type of change

- [x] Breaking change (fix/feature causing existing functionality to break)

# Additional information

- Python version: 3.9-3.12
- OpticStudio version: 24.1

## Related issues

Closes #15 

## Checklist

    - New tests will be added after evaluating this proposal.


zospy/analyses/new/base.py
```diff
@@ -0,0 +1,529 @@
...
+
+    @staticmethod
+    def _create_tempfile(path: Path | None, suffix: str) -> (Path, bool):
+        if path is None:
+            fd, path = mkstemp(suffix=suffix, prefix="zospy_")
+            os.close(fd)
+            clean_file = True
+        else:
+            if not path.suffix == suffix:
+                raise ValueError(f"cfgfile should end with '{suffix}'")
```
LucVV, line 474:
```suggestion
                raise ValueError(f"filepath should end with '{suffix}'")
```

zospy/analyses/new/base.py
```diff
@@ -0,0 +1,529 @@
...
+            return
+        else:
+            raise ValueError(f"oncomplete should be a member of zospy.analyses.base.OnComplete, got {self.oncomplete}")
+
+    def run(self, oss: OpticStudioSystem, *args, **kwargs) -> AnalysisResult[AnalysisData, AnalysisSettings]:
+        self._oss = weakref.proxy(oss)
+        self._create_analysis()
+
+        self.config_file, self._remove_config_file = self._create_tempfile(self.config_file, ".CFG")
+        self.text_output_file, self._remove_text_output_file = self._create_tempfile(self._text_output_file, ".txt")
```
LucVV, line 498: If I understand correctly, we are now always making the temporary files, even in instances where they are not needed?

zospy/analyses/new/raysandspots/single_ray_trace.py
```diff
@@ -0,0 +1,150 @@
...
+from pydantic.dataclasses import dataclass
+
+from zospy.analyses.new.base import AnalysisWrapper
+from zospy.analyses.new.parsers.transformers import ZospyTransformer
+from zospy.analyses.new.parsers.types import UnitField
+from zospy.api import constants
+
+
+class SingleRayTraceTransformer(ZospyTransformer):
+    def NAN(self, args):
```
LucVV, line 18: This does not seem to be used?

crnh: `NAN` is defined in `single_ray_trace.lark`, so it is used. You can read how in Lark's documentation on transformers: https://lark-parser.readthedocs.io/en/stable/visitors.html#transformer
Summary: the methods of transformers must have the same name as the tokens defined in the grammar.

zospy/analyses/new/base.py
crnh: Yes, because we need them in most cases. Alternatively, I can add two boolean class attributes (e.g. `needs_text_output_file` and `needs_config_file`) to create the files only when needed.

crnh:
> If I see correctly, we will use `.GetTextFile()` for all analyses and parse that file, or did I miss another way for analyses for whom the results are directly available in the API, such as the current `huygens_psf()`

This PR is mainly a proof of concept for the new parsers, so I did not yet implement analyses whose results can be accessed through the API. Accessing results throughout the API is preferred whenever possible, but methods for that still need to be added.

> With regards to the design considerations:
> 
>     * I think adding the analysis settings to the result object is important. That way, they can be saved with the result, and one can always see what settings were used for a specific analysis without having to dive into source code.
> 
>     * With regard to the analyses with many parameters, I am unsure what would be optimal. My thourhgts:
>       
>       * For ease of use, it would be good to retain the functionality of e.g. IntelliSense.
>       * By specifying the default values, we also have clear defaults for ZOSPy. In interactive mode in OpticStudio, one can adjust de default configuration file, therewith adjusting the new default values. I do not exactly know how this interacts with the API. It would suboptimal if the same lines of code can yield different output on different systems due to locally saved defaults.
>       * I agree that asymmetry between analyses would be weird and should be omitted.
> 
>     * Passing config file, text output file, and oncomplete to `run` sounds good. They do not have to be in the analysis result per se in my opinion.

I discussed this with @jwmbeenakker, and we decided to

- Include the analysis settings in the result;
- Implement something in-between the current situation and option ii, and using unit tests to ensure that the default parameters of the analyses' `__init__` methods match the defaults of the settings objects;
- Pass the config file, text output file and oncomplete to the `run` method.

LucVV:
This is a neat solution! If I understand correctly, I also allows you to create a separate `Settings()` instance and pass that to settings right? That might be insightfull for PhysicalOpticsPropagation and such.

zospy/analyses/new/base.py
```diff
@@ -0,0 +1,560 @@
...
+        self._config_file = None
+        self._text_output_file = None
+
+        self._oss = None
+        self._analysis = None
+        self._remove_config_file = False
+        self._remove_text_output_file = False
+
+    def _init_settings(self, settings: AnalysisSettings, parameters: dict[str, any]):
+        parameters.pop("self", None)  # Remove self from parameters if it exists
```
LucVV, line 402: In the analyses, we now use:
```python
    def __init__(
        self,
        sampling: str | int = "32x32",
        ...,
        settings: PolarizationTransmissionSettings | None = None,
    ):
        super().__init__(settings or PolarizationTransmissionSettings(), locals())
```

I think `settings` are also in `locals()` in this case. Should we therefore not change this to:
```suggestion
        parameters.pop("self", None)  # Remove self from parameters if it exists
        parameters.pop("settings", None)  # Remove settings from parameters if it exists
```

crnh: You're right, I added the `settings` parameter after writing the `_init_settings` method, and didn't update the latter afterwards. I made the method more generic now, so instead of popping some variables, it iterates through the fields of the setting class to update them.

crnh:
> This is a neat solution! If I understand correctly, I also allows you to create a separate `Settings()` instance and pass that to settings right? That might be insightfull for PhysicalOpticsPropagation and such.

Yes, it should indeed be possible to create a settings instance and pass that to the analysis constructor.

zospy/analyses/new/base.py
LucVV: I like this implementation! We should be wary that we don't implement **kwargs in any analysis class, as then one could specify e.g. `hx=1` rather than `Hx=1` and it would go unnoticed. But since we do not plan on doing so, maybe we can ignore that. If you agree, we can resolve this conversation

LucVV:
All tests now pass for me!

jwmbeenakker (changes requested):
Great work. Will help a lot for future inclusion of ZOSAPI functions. Could you write a 1 or 2 sentence description in the changelog, so we keep track of when new functions have been added

jwmbeenakker (approved):
If you agree with my small addition to the changelog, this PR is good to go

jwmbeenakker (approved):

crnh:
I implemented JSON serialization and deserialization for analysis results. The current implementation is quite similar to the old implementation, but taking advantage from Pydantic's JSON conversion features for dataclasses. This was necessary because the `AnalysisResult` class is a generic class, which can take any type for the data and settings objects.

The conversion process would be a lot easier if the analysis data and settings types were constrained, but this requires `zospy.analyses.new.base` to import the data and settings types from the analyses. The analyses in turn import `AnalysisWrapper` from `zospy.analyses.new.base`, resulting in a circular import. This can only be resolved by splitting the analysis data and result objects to separate files, and I don't know if we want that.

jwmbeenakker (approved):

LucVV:
All tests pass now after this change!

PR #80, opened 2024-10-22
Add support for OpticStudio viewer exports

crnh:
## Proposed change

Add support for the sequential 3D viewer and cross section export tools. Notebooks that currently contain a screenshot of the optical system have been updated to use these viewers.

I only implemented these two viewers and did not add unit tests; adding and testing new viewers will be significantly more efficient when the new analysis structure is implemented, so the tests and potentially other viewers will be implemented after that.

## Type of change

# Additional information

- Python version: 3.9-3.12
- OpticStudio version: 24R1

## Checklist


jwmbeenakker (changes requested):
Nice work. Small suggestion: add a system viewer plot to the "full example" of the README 
https://zospy--80.org.readthedocs.build/en/80/index.html#full-example

crnh:
> Nice work. Small suggestion: add a system viewer plot to the "full example" of the README
> https://zospy--80.org.readthedocs.build/en/80/index.html#full-example

That requires an image to be added to the README, i.e. another file littering the project root. Do we really want that?

LucVV:
I think this is a great improvement! The only thing I am worried about is that it "breaks" the systemviewer analyses for any version <2024. In those versions, `Tools` does not have the `Layouts` attribute in the API. The old implementation was kind of bare, so it is not a big issue, but you will also need version 2024 to run the examples. We can agree on this, but it has to be a conscious choice. As an alternative, we can either implement a 'legacy" version of the function, or make sure that the function raises a warning and returns none for older OpticStudio versions.

jwmbeenakker:
> The only thing I am worried about is that it "breaks" the systemviewer analyses for any version <2024. 

Good point (which we will likely encounter more often as the features of OpticStudio expand). Maybe add a check if the OS-version >=24
(maybe it is a good idea to always add a minimum version to analyses, now we are rewriting the whole analysis part #78 ?)

crnh:
> I think this is a great improvement! The only thing I am worried about is that it "breaks" the systemviewer analyses for any version <2024

I already planned a check for that, but didn't push it yet. It will check if OS < 24 and in that case return no data (but open the viewer anyways).

> (maybe it is a good idea to always add a minimum version to analyses, now we are rewriting the whole analysis part #78 ?)

No, because for most analyses we do not know they were added to OpticStudio, so adding this check will definitely break things for users. It may be an acceptable solution if some analyses appear not to work for older OpticStudio versions, but then I prefer to only add this for analyses which users reported to not work for older versions.

crnh:
@LucVV can you check if you can still run the analyses on your system?

jwmbeenakker (approved):

LucVV:
The analyses now work again on my system, so this seems like a good solution. I have two points however:
1. In any version <24, all parameters that are specified are ignored, as they should. However, should we notify users about this using a (suppressible) waring?
2. In the examples, we now call `plt.imshow(draw3d.Data)`, which breaks on versions <24 because `draw3d.Data` is `None`. As a result, the examples are somewhat broken there.

crnh:
> 1. In any version <24, all parameters that are specified are ignored, as they should. However, should we notify users about this using a (suppressible) waring?

Checking for specified parameters is not straightforward because of the default values, but I think there is a way to do this. I will create a second PR for this.

> 2\. In the examples, we now call `plt.imshow(draw3d.Data)`, which breaks on versions <24 because `draw3d.Data` is `None`. As a result, the examples are somewhat broken there.

I understand this, but choosing to not demonstrate features that are only available in recent OpticStudio versions makes no sense. I will add a warning to the notebook.

PR #82, opened 2024-10-28
Move from Tox to Hatch

crnh:
## Proposed change

We currently use Tox together with a bunch of other tools to perform common tasks on ZOSPy, such as linting and formatting.
Changing to Hatch (https://hatch.pypa.io/), a project manager endorsed by the Python Packaging Authority, allows to create a more unified workflow.

Some advantages over Tox that are relevant for ZOSPy:

- Automated management of Python versions: if a version required by the unit tests is not installed, Hatch will install it when setting up the test environment.
- Unified linting and formatting: we currently use Black, Isort and Ruff, but Ruff can nowadays perform linting, formatting and import sorting. Hatch has built-in support for Ruff (`hatch fmt`).
- Configuration of both project and tool settings in `pyproject.toml` (so no need anymore for a separate `tox.ini` file)

The current implementation also provides some additional commands:

- `hatch run docs` to build and preview the documentation;
- `hatch run generate-reference-data` to generate test reference data.

The current Ruff configuration highlighted a bunch of minor issues that I will try to fix (or ignore) as part of this PR. Some changes will be breaking, so this should be done between ZOSPy 1.3.0 and ZOSPy 2.0.0.

Additional changes:

- Added a script to run all notebooks. Notebooks are run in separate isolated environments.
- Migrated the workflows to hatch + uv.

## Type of change

- [ ] Example (a notebook demonstrating how to use ZOSPy for a specific application)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New analysis (a wrapper around an OpticStudio analysis)
- [ ] New feature (other than an analysis)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests
- [ ] Documentation (improvement of either the docstrings or the documentation website)

# Additional information

- Python version: 3.12
- OpticStudio version: 24 R1.03

## Checklist

- [x] I have followed the [contribution guidelines][contribution-guidelines]
- [x] The code has been linted, formatted and tested locally using tox.
- [ ] Local tests pass. **Please fix any problems before opening a PR. If this is not possible, specify what doesn't work and why you can't fix it.**
- [ ] I added new tests for any features contributed, or updated existing tests.
- [ ] I updated CHANGELOG.md with my changes (except for refactorings and changes in the documentation).

If you contributed an analysis:

- [ ] I did not use `AttrDict`s for the analysis result data (please use dataclasses instead).

If you contributed an example:

[...]

PR #88, opened 2024-11-01
Apply suggestions from Ruff

crnh:
## Proposed change

This is a follow-up PR for #82, applying the refactorings suggested by Ruff. 
Furthermore, I noticed that `rgetattr` in `zospy.utils.pyutils` is also available in Python's standard library, so I replaced it with the standard library function.

This introduces some breaking changes, as Ruff suggests to make boolean arguments keyword-only. In my opinion it makes sense to follow this suggestion, because it improves code readability, e.g.:

```python
# No clue what is True here
zos = zp.ZOS(True)

# This is better
zos = zp.ZOS(preload=True)
```

## Type of change

# Additional information

- Python version: 3.9-3.12
- OpticStudio version: 24 R1.3

## Checklist


jwmbeenakker (commented):
Looks good. Did you rerun all the jupyter notebooks, to verify they still run, as they are (I think) not yet part of the unit tests?

jwmbeenakker:
> > not yet part of the unit tests
> 
> We don't plan to make them part of the unit tests, right?

Yes, would be great if they become part of these test. 
But to finish this PR: could you (manually?) check if they run (as quite some of the variables have been renamed), so we are sure people can use them as examples?

crnh:
> Yes, would be great if they become part of these test.

They are not unit tests and do not fit in the definition of unit tests. If the unit tests pass, these notebooks should be able to run as well. Some examples are rather time-consuming, so automatically running them each time when I run the unit tests will soon become a nuisance. Including them in the unit tests will also increase the risk of running them on older versions of OpticStudio, which may result in incomplete examples with warnings being included in the documentation.

Rerunning the notebooks when they have changed or prior to a release makes sense, of course, but I think the current setup automates that sufficiently. 

> But to finish this PR: could you (manually?) check if they run

See my previous comment.

jwmbeenakker (approved):

PR #97, opened 2024-12-13
New-style system viewers

crnh:
## Proposed change

Convert system viewers to the new format. The Cross Section and 3D Layout viewers are fully implemented. The other viewers support exporting images, but do not (yet) allow to change settings.

## Type of change

- [ ] Example (a notebook demonstrating how to use ZOSPy for a specific application)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New analysis (a wrapper around an OpticStudio analysis)
- [ ] New feature (other than an analysis)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation (improvement of either the docstrings or the documentation website)

# Additional information

- Python version: 3.9-3.12
- OpticStudio version: 24 R2

## Checklist

- [x] I have followed the [contribution guidelines][contribution-guidelines]
- [x] The code has been linted, formatted and tested locally using tox.
- [ ] Local tests pass. **Please fix any problems before opening a PR. If this is not possible, specify what doesn't work and why you can't fix it.**
- [x] I added new tests for any features contributed, or updated existing tests.
- [ ] I updated CHANGELOG.md with my changes (except for refactorings and changes in the documentation).

If you contributed an analysis:

- [x] I did not use `AttrDict`s for the analysis result data (please use dataclasses instead).

If you contributed an example:

- [ ] I contributed my example as a Jupyter notebook.
    

[contribution-guidelines]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/CONTRIBUTING.md
[unittest-instructions]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/tests/README.md
[numpydoc]: https://numpydoc.readthedocs.io/en/latest/format.html

PR #100, opened 2024-12-16
Implemented a fix for instances where datagrid.Description did not su…

LucVV:
…pply the full description

## Proposed change
  The new curvature analysis uses `datagrid.Description` to obtain information on the centration of the grid. `datagrid.Description` however returns different information in older OpticStudio versions, causing an error in the curvature analysis. This PR fixes that issue with aa fallback to `GetTextFile` if `datagrid.Description` provides insufficient information.

## Type of change

- [ ] Example (a notebook demonstrating how to use ZOSPy for a specific application)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New analysis (a wrapper around an OpticStudio analysis)
- [ ] New feature (other than an analysis)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation (improvement of either the docstrings or the documentation website)

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Related issues

#94 

## Checklist

- [x] I have followed the [contribution guidelines][contribution-guidelines]
- [x] The code has been linted, formatted and tested locally using tox.
- [x] Local tests pass. **Please fix any problems before opening a PR. If this is not possible, specify what doesn't work and why you can't fix it.**
- [ ] I added new tests for any features contributed, or updated existing tests.
- [ ] I updated CHANGELOG.md with my changes (except for refactorings and changes in the documentation).

If you contributed an analysis:

- [ ] I did not use `AttrDict`s for the analysis result data (please use dataclasses instead).

If you contributed an example:

- [ ] I contributed my example as a Jupyter notebook.
    

[contribution-guidelines]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/CONTRIBUTING.md
[unittest-instructions]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/tests/README.md
[numpydoc]: https://numpydoc.readthedocs.io/en/latest/format.html

PR #101, opened 2024-12-17
Systemviewers fixes for older OpticStudio versions

LucVV:
## Proposed change

The new systemviewer implementation does not fully work with older versions of OpticStudio. There are two main issues:
1. The SystemViewer analyses do not return any result with older OpticStudio versions, making `AnalysisData` `None`. In principlle this is oke, but it breaks the current json serialization. This is mainly a problem for some unit tests.
2. The unit tests also check if the SystemViewer analysis works using `assert result.AnalysisData is not None`, which fails for older OpticStudio versions due to point 1.

## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Related issues

#97 

## Checklist

### DIFF
--- a/tests/analyses/new/test_systemviewers.py
+++ b/tests/analyses/new/test_systemviewers.py
@@ -14,10 +14,11 @@ from zospy.analyses.new.systemviewers.base import SystemViewerWrapper
 
 def consider_minimal_opticstudio_version_in_result(result, minimal_version):
     """Makes sure systemviewer results is correctly asserted for different versions of OpticStudio."""
-    if minimal_version >= '24.1.0':
+    if minimal_version >= "24.1.0":
         assert result.data is not None
     else:  # No result.data as layout exports are not supported
         assert result.data is None
+
 
 class TestBase:
     @analysis_settings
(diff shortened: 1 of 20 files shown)


LucVV:
Similar as in #100, I think changelog can be skipped for this one?

crnh:
> Similar as in #100, I think changelog can be skipped for this one?

Yes, this is still part of the analysis conversion, which will be listed as a single item in the changelog.

tests/analyses/new/test_systemviewers.py
```diff
@@ -12,6 +12,14 @@
 from zospy.analyses.new.systemviewers.base import SystemViewerWrapper
 
 
+def consider_minimal_opticstudio_version_in_result(result, minimal_version):
```
crnh, line 15: I'd choose a name that shows that the assert happens in this function, e.g. `assert_systemviewer_result`

crnh (changes requested):

tests/analyses/new/test_systemviewers.py
LucVV: Name is changed to a more concise one!

crnh (approved):

PR #103, opened 2024-12-18
Obtain correct origin for the geometric image analysis

LucVV:
## Proposed change

As discussed on the [OpticStudio forum](https://community.zemax.com/zos-api-12/incorrect-datagrid-miny-for-geometric-image-analysis-5426), the origin of the GeometricImageAnalysis is incorrect. It was suggested to calculate the origin from the reference, which is implemented in this PR. 

There is still a question that needs to be answered: when `wavelength` is set to "All",  `reference` to anaything but "PrimaryChief" and there are multiple wavelengths, is the OpticStudio reference the average of the references calculated with each wavelength? This question is asked on the OpticStudio forum as the manual does not provide an answer.

I have currently implemented this fix for the old version of the Geometric Image Analysis, especially since it is still to be converted to the new format. 

Due to the above reasons, this PR is set as draft.
 
## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist


crnh:
This seems to be pretty complicated, I think it's best to add some unit tests for this as well...

LucVV:
I will add them soon!

crnh:
This should still be implemented for the new-style analysis!

crnh:
Two comments:

1. `_gia_correct_datagrid_origin` does not seem to be used anywhere;
2. In my opinion, using a wrapper class to fix this problem is unnecessarily complex. Instead, modifying the `unpack_datagrid` function to allow for optionally changing the minimum x and y values would work equally well and be a lot easier to implement, test and maintain.

LucVV:
This fix was actually much easier than expected. The origin of the datagrid is `-Settings.ImageSize/2`. I have implemented this for both `MinX` and `MinY` and added a test for it. The only thing I am unsure about is what happens to the output when `image_size` is lower than 0. The OpticStudio user interface seems to change its output based on what you select as `ShowAs`. I am not sure if we should limit `image_size` to >=0, or do something else. @crnh do you have an opinion on this?

crnh:
> I am not sure if we should limit `image_size` to >=0, or do something else. @crnh do you have an opinion on this?

As the behavior for negative image sizes is rather weird / unpredictable, I would be okay with not supporting negative image sizes. If later on they appear to be useful, it's easy enough to add support for them.

zospy/analyses/old/extendedscene.py
```diff
@@ -8,9 +8,133 @@
 
 from zospy import utils
 from zospy.analyses.old.base import AnalysisResult, OnComplete, new_analysis
-from zospy.api import constants
+from zospy.api import constants, _ZOSAPI
 from zospy.zpcore import OpticStudioSystem
 
+class _GiaDataGridWrapper:
```
crnh, line 14: Is this wrapper still necessary? If not, I think it can be removed.

crnh (changes requested):
The test reference data still needs to be updated. I'll do that for OpticStudio 2024.

crnh:
I ran the tests for OpticStudio 2024 R1 and R2, and they pass!

zospy/analyses/old/extendedscene.py
LucVV: You are correct, I removed all changes from `analyses/old/extendescene`

crnh (approved):
Nice work! Thanks for investigating and fixing this. For future reference, this behavior has been discussed on the OpticStudio forums: https://community.zemax.com/zos-api-12/incorrect-datagrid-miny-for-geometric-image-analysis-5426?postid=17836#post17836

PR #110, opened 2025-01-13
Refactor the way analysis settings are set

crnh:
## Proposed change

This is a follow-up on #108.

In the current implementation, analysis settings can be specified using both a settings object and keyword arguments.
In some cases, this implementation can prevent changing modified settings back to their default values.
Implementations that don't exhibit this behavior rely on `**kwargs`, which means we lose the specification of the default arguments in the function signature.

The `__init__` method of analysis wrappers now only allows to specify the settings using keyword arguments. Initializing an analysis with a settings object can be done using the `from_settings` classmethod, e.g. `ZernikeStandardCoefficients.from_settings(ZernikeStandardCoefficientsSettings(sampling="64x64", maximum_term=45))`. **A downside of this implementation is that we lose the ability to create a new 'default' set of settings and then easily overwrite some of these settings**, which was a major reason to choose the previously proposed implementation, but we can probably live with that.

Other (related) changes:

- The type of the analysis settings object is now determined from the class signature, e.g. `class PolarizationTransmission(
    BaseAnalysisWrapper[PolarizationTransmissionResult, PolarizationTransmissionSettings]): ...` is sufficient to let `BaseAnalysisWrapper` find out that settings should have the `PolarizationTransmissionSettings` type. As a result, it is no longer necessary to create a `PolarizationTransmissionSettings` instance in `PolarizationTransmission.__init__`.
- All analysis classes contained some configuration parameters (e.g. the analysis type) that were specified as class attributes. These are now specified as arguments instead. See the example below:

```python
# Old specification
class PolarizationTransmission(
    BaseAnalysisWrapper[PolarizationTransmissionResult, PolarizationTransmissionSettings],
):
    TYPE = "Transmission"
    _needs_config_file = True
    _needs_text_output_file = True

    ...

# New specification
class PolarizationTransmission(
    BaseAnalysisWrapper[PolarizationTransmissionResult, PolarizationTransmissionSettings],
    analysis_type="Transmission",
    needs_config_file=True,
    needs_text_output_file=True,
):
    ...
```

## Type of change

# Additional information

- Python version: 3.9-3.12
- OpticStudio version: 24R2

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,8 @@ path = "zospy/__init__.py"
 python = "3.12"
 installer = "uv"
 path = ".venv"
+dependencies = ["pytest"]
+
 
 [tool.hatch.envs.default.scripts]
 test-extension = "hatch test --extension {args}"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,17 +5,17 @@ build-backend = "hatchling.build"
 [project]
 name = "zospy"
 authors = [
-    {name = "Luc van Vught"},
-    {name = "Jan-Willem Beenakker"},
-    {name = "Corné Haasjes"}
+    { name = "Luc van Vught" },
+    { name = "Jan-Willem Beenakker" },
+    { name = "Corné Haasjes" }
 ]
 maintainers = [
-    {name = "MReye research group", email = "zospy@mreye.nl"}
+    { name = "MReye research group", email = "zospy@mreye.nl" }
 ]
 
 description = "A Python package used to communicate with Zemax OpticStudio through the API"
 readme = "README.md"
-license = {file = "LICENSE.txt"}
+license = { file = "LICENSE.txt" }
 keywords = ["Zemax", "OpticStudio", "API", "ZOSAPI"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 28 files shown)


jwmbeenakker (approved):

crnh:
I added a new `update_settings` method to allow for batch-updating settings on an analysis using values from a dictionary. This method was already used internally to change the settings, so effectively this doesn't change much.

LucVV:
> The classmethod is now called `from_settings` but we do not really create an analysis wrapper _from_ those settings, but rather _with_ those settings. `with_settings` may therefore be a better name. @LucVV which one do you prefer?

I think `with_settings` is better

LucVV (commented):
Looks good! We should however pull v2.0.0 and update the analyses of #112 and #114 to implement these changes

LucVV:
> We should however pull v2.0.0 and update the analyses of #112 and #114 to implement these changes

I implemented this and checked (and corrected) all other tests. Now we only have to decide on `from_settings`/`with_settings`

crnh:
> Now we only have to decide on `from_settings`/`with_settings`

I already changed `from_settings` to `with_settings`.

LucVV:
Then this PR is good to go in my opinion! It might be good if you run the tests one more time after my (small) changes

LucVV (approved):

PR #115, opened 2025-02-07
Implemented skip for Tools.Layouts in test_constant_exists with older OpticStudio versions

LucVV:
## Proposed change

`test_constant_exists` currently tests all constants, including constants under `Tools.Layouts`. This fails on OpticStudio versions < 24R1 as `Tools.Layouts` did not exist yet. I think it is optimal to implement the skip within the tests in this instance, as the tested constants are outomatically determined.

## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist

### DIFF
--- a/tests/analyses/new/parsers/test_types.py
+++ b/tests/analyses/new/parsers/test_types.py
@@ -7,7 +7,6 @@ from pandas import DataFrame
 from pandas.testing import assert_frame_equal
 from pydantic import TypeAdapter, ValidationError
 
-from tests.conftest import optic_studio_version
 from zospy.analyses.new.parsers.types import ValidatedDataFrame, ValidatedNDArray, ZOSAPIConstantAnnotation
 from zospy.api import constants
 
--- a/tests/analyses/new/parsers/test_types.py
+++ b/tests/analyses/new/parsers/test_types.py
@@ -7,6 +7,7 @@ from pandas import DataFrame
 from pandas.testing import assert_frame_equal
 from pydantic import TypeAdapter, ValidationError
 
+from tests.conftest import optic_studio_version
 from zospy.analyses.new.parsers.types import ValidatedDataFrame, ValidatedNDArray, ZOSAPIConstantAnnotation
 from zospy.api import constants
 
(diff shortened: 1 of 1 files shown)


crnh (commented):
Thanks for fixing this! I expect we will see more constants being added in future releases, so we may consider a more generic solution (e.g. with a dictionary mapping constants to the version in which they were added). Do you think we should opt for a more generic implementation now, or should we wait until more constants are added?

LucVV:
> Thanks for fixing this! I expect we will see more constants being added in future releases, so we may consider a more generic solution (e.g. with a dictionary mapping constants to the version in which they were added). Do you think we should opt for a more generic implementation now, or should we wait until more constants are added?

It might be nice to directly make it future proof. I see two options:

```python
MINIMAL_ZOS_VERSION_FOR_CONSTANT = {  # Supply as {"constant": "zos version string"}
    "Tools.Layouts": "24.1.0"
}

# ...

    def test_constant_exists(self, zos, annotation, optic_studio_version):  # noqa: ARG002
        for constant_name, minimal_optic_studio_version in MINIMAL_ZOS_VERSION_FOR_CONSTANT.items():
            if annotation.enum.startswith(constant_name) and optic_studio_version < minimal_optic_studio_version:
                pytest.skip(f"{annotation.enum} is only available from OpticStudio {minimal_optic_studio_version} onwards")
```

which is quite inefficient, but does the trick. Or

```python
MINIMAL_ZOS_VERSION_FOR_CONSTANTS = {  # Supply as {"zos version string": constant}
    "24.1.0": ["Tools.Layouts"]
}

# ...
    
    def test_constant_exists(self, zos, annotation, optic_studio_version):  # noqa: ARG002
        for minimal_optic_studio_version, constant_names in MINIMAL_ZOS_VERSION_FOR_CONSTANTS.items():
            if optic_studio_version < minimal_optic_studio_version and any(annotation.enum.startswith(constant_name) for constant_name in constant_names):
                pytest.skip(f"{annotation.enum} is only available from OpticStudio {minimal_optic_studio_version} onwards")
```

Which might be slightly more efficient as we first evaluate if we need to check it for this optic_studio_version and only if so we loop through the constants that should be skipped. Was this what you had in mind?

LucVV:
Works great!

PR #116, opened 2025-02-10
Added method to convert Zernike standard coefficients to dataframe

LucVV:
## Proposed change

I added a  method to quickly obtain the coefficients from `ZernikeStandardCoefficientResult` to a pandas `DataFrame`. As `ZernikeStandardCoefficientResult.coefficients` is a dictionary and embedding that in a Pydantic `RootModel` while maintaining full dictionary functionality is not straightforward, it is added as property to `ZernikeStandardCoefficientResult`: `ZernikeStandardCoefficientResult.coefficients_dataframe`.

## Type of change

- [ ] Example (a notebook demonstrating how to use ZOSPy for a specific application)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New analysis (a wrapper around an OpticStudio analysis)
- [x] New feature (other than an analysis)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation (improvement of either the docstrings or the documentation website)

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist

- [x] I have followed the [contribution guidelines][contribution-guidelines]
- [x] The code has been linted, formatted and tested locally using tox.
- [x] Local tests pass. **Please fix any problems before opening a PR. If this is not possible, specify what doesn't work and why you can't fix it.**
- [ ] I added new tests for any features contributed, or updated existing tests.
- [ ] I updated CHANGELOG.md with my changes (except for refactorings and changes in the documentation).

If you contributed an analysis:

- [ ] I did not use `AttrDict`s for the analysis result data (please use dataclasses instead).

If you contributed an example:

- [ ] I contributed my example as a Jupyter notebook.
    

[contribution-guidelines]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/CONTRIBUTING.md
[unittest-instructions]: https://github.com/MREYE-LUMC/ZOSPy/blob/main/tests/README.md
[numpydoc]: https://numpydoc.readthedocs.io/en/latest/format.html

PR #117, opened 2025-02-11
Improved AnalysisResult for FFTThroughFocusMTF

LucVV:
## Proposed change

I implemented `FFTThroughFocusMTFData`, `FFTThroughFocusMTFResult` and a `to_dataframe` method for the `FFTThroughFocusMTF` analysis. It will now keep the results for different fields separate per field, and if wanted a long DataFrame can be created with all coordinates, similar to the `RayFan` analysis.

For the `FFTThroughFocusMTFResult`, I chose to subclass a pydantic `RootModel`. Another option would be subclassing a list:

```python
@analysis_result
class FFTThroughFocusMTFResult(list[FFTThroughFocusMTFData]):
    """Data for the Ray Fan analysis.

    Methods
    -------
    to_dataframe()
        Convert the data to a single Pandas DataFrame.
    """

    def to_dataframe(self) -> DataFrame:
        """Convert the data to a Pandas DataFrame.

        The separate dataframes for each field are combined in a DataFrame in long format.
        In addition to the columns for each wavelength, the returned DataFrame has the following columns:

        - FieldX: The field x coordinate
        - FieldY: The field y coordinate

        Returns
        -------
        DataFrame
            The data in long format.
        """
        dataframes = []

        for fft in self:
            df = fft.to_dataframe()
            dataframes.append(df)  # Move the index to a separate column to prevent overlap

        return pd.concat(dataframes, ignore_index=True).reset_index(drop=True)
``` 

but that might be not transferable to other implementations. For example, in #116, we can use a `RootModel` with a dict root, but directly subclassing dict instead break validations.

One remaining question is, if we have one field and thus one analysis return, would we want to set `AnalysisResult.data` to a `FFTThrougFocusAnalysisResult` with length 1, or rather directly to the single `FFTThrougFocusAnalysisData` instance?

## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist


crnh:
This will be considered after the release of ZOSPy 2.0.

zospy/analyses/new/mtf/fft_through_focus_mtf.py
```diff
@@ -58,8 +62,65 @@ class FFTThroughFocusMTFSettings:
...
+        DataFrame
+            The data in long format.
+        """
+        dataframes = []
+
+        for fft in self:
+            df = fft.to_dataframe()
+            dataframes.append(df)  # Move the index to a separate column to prevent overlap
+
+        return pd.concat(dataframes, ignore_index=True).reset_index(drop=True)
```
crnh, line 119: You can use a list comprehension here:
```suggestion
        return pd.concat([fft.to_dataframe() for fft in self], ignore_index=True).reset_index(drop=True)
```

And reduce the method to a single line

crnh (changes requested):
Does the serialization work? I think custom root models are not supported out of the box

zospy/analyses/new/mtf/fft_through_focus_mtf.py
LucVV: You were correct, I ran the wrong `test_mtf` to test. I have now implemented a pydantic basemodel (de)serialization, now all should work

crnh:
Tests pass on my system as well. This PR should be merged after #118, as I am currently validating the analysis results using the old data structure for this analysis. After the validation, we can merge this and generate new reference data for all analyses.

LucVV:
I updated this PR as requested in https://github.com/MREYE-LUMC/ZOSPy/issues/123, after the folder structure for v2.0.0 was changed. I do need new reference data for `TestFFTThroughFocusMTF` to check if all is correct. @crnh can you create that?

crnh:
I updated the reference data for FFTThroughFocusMTF. I also had to update the reference data for all other tests, because we renamed the data_type 'dataclass' to 'zospy_class'. All tests now pass on my system.

crnh (approved):

PR #121, opened 2025-02-19
Generate new reference data for the new-style analyses

crnh:
## Proposed change

Generate new reference data for the new-style analyses for OpticStudio 2024R1 and 2024R2.
No data was generated for the system viewers, as these are only used for visualization purposes, but I can still add tests for data returned by those analyses as well.

## Type of change

# Additional information

- Python version: 3.12
- OpticStudio version: 2024R1, 2024R2

## Checklist


jwmbeenakker:
Thank you for this invisible, but important, work. 
For future reference:
you also checked this v2 output against the v1 output. Can you specify it there were any (significant) differences?

crnh:
> you also checked this v2 output against the v1 output. Can you specify it there were any (significant) differences?

That's already mentioned in #120 .

crnh:
image
I see the results for the Shack-Hartmann examples are different now. @LucVV as you made these examples, can you check what causes the differences?

crnh:
> I see the results for the Shack-Hartmann examples are different now. @LucVV as you made these examples, can you check what causes the differences?

This is likely caused by randomness in the Geometric Image Analysis.

LucVV:
@crnh sorry for the delayed reply. That is correct, I remember extensively investigating why my example did not match the knowledgebase example, untill I realized that the output is indeed altering slightly due too randomness

PR #125, opened 2025-03-03
Functions for setting/getting field, wavelength and surface

LucVV:
## Proposed change

We currently use properties and setters to adjust the field and wavelength for an analysis. They had to be implemented directly on the analysis class rather than in `Analysis.Settings`. This makes it a bit unintuitive what these properties do, they change a setting but are not under `.Settings`. To make it more obvious, this PR changes them to dedicated `get_` and `set_`functions.

Furthermore, the `set_surface` for the new style analyses does not seem to have the fix that we implemented in the old version of `set_surface`, the current version only allows `'Objective'`, while some analyses might parse `'Object'` instead. I implemented that also in this PR. 

## Type of change

- [x] Breaking change (fix/feature causing existing functionality to break)

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist

### DIFF
--- a/zospy/analyses/base.py
+++ b/zospy/analyses/base.py
@@ -342,7 +342,7 @@ class Analysis:
         Parameters
         ----------
         value : int | Literal["All"]
-            The value to which the field should be set. Either int or str. Accepts only 'All' as string.
+            The value to which the field should be set. Either int or str. When value == 0, all fields are used. Accepts only 'All' as string.
 
         Returns
         -------
@@ -352,10 +352,9 @@ class Analysis:
         ------
         ValueError
             When 'value' is not integer or string. When it is a string, it also raises an error when the string
-            does not
-            equal 'All'.
+            does not equal 'All'.
         """
-        if value == "All":
+        if value == "All" or value == 0:
             self.Settings.Field.UseAllFields()
         elif isinstance(value, int):
             self.Settings.Field.SetFieldNumber(value)
@@ -427,7 +426,7 @@ class Analysis:
         Parameters
         ----------
         value : int | Literal["All"]
-            The value to which the wavelength should be set. Either int or str. Accepts only 'All' as string.
+            The value to which the wavelength should be set. Either int or str. When value == 0, all wavelengths are used. Accepts only 'All' as string.
 
         Returns
         -------
@@ -439,7 +438,7 @@ class Analysis:
             When 'value' is not integer or string. When it is a string, it also raises an error when the string does not
             equal 'All'.
         """
-        if value == "All":
+        if value == "All" or value == 0:
             self.Settings.Wavelength.UseAllWavelengths()
         elif isinstance(value, int):
             self.Settings.Wavelength.SetWavelengthNumber(value)
--- a/zospy/analyses/base.py
+++ b/zospy/analyses/base.py
@@ -342,7 +342,7 @@ class Analysis:
         Parameters
         ----------
         value : int | Literal["All"]
-            The value to which the field should be set. Either int or str. When value == 0, all fields are used. Accepts only 'All' as string.
+            The value to which the field should be set. Either int or str. Accepts only 'All' as string.
 
         Returns
         -------
@@ -426,7 +426,7 @@ class Analysis:
         Parameters
         ----------
         value : int | Literal["All"]
-            The value to which the wavelength should be set. Either int or str. When value == 0, all wavelengths are used. Accepts only 'All' as string.
+            The value to which the wavelength should be set. Either int or str. Accepts only 'All' as string.
 
         Returns
         -------
--- a/zospy/analyses/base.py
+++ b/zospy/analyses/base.py
@@ -363,7 +363,7 @@ class Analysis:
             raise ValueError(f'Field value should be "All" or an integer, got {value}')
 
         if message is not None:
-            raise ValueError(f'Could not set field value to {value}: {message.Text}')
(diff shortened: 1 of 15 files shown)


crnh:
> Not sure if we want to skip changelog for this PR as it is in a way a breaking change.

The behavior is the same as in ZOSPy 1.3.1. The changelog will list the changes between 1.3.1 and 2.0.0, so no need to add this to the changelog, in my opinion.

zospy/analyses/base.py
```diff
@@ -324,9 +324,8 @@ def Results(self) -> _ZOSAPI.Analysis.Data.IAR_:  # noqa: N802
         """Analysis results."""
         return self._analysis.GetResults()
 
-    @property
-    def field(self) -> int | str:
-        """Gets the wavelength value of the analysis.
+    def get_field(self) -> int | str:
+        """Get the field value of the analysis.
```
crnh, line 328:
```suggestion
        """Get the field value from the analysis settings.
```
The docstrings can state more explicitly that these methods access / modify the analysis settings.

zospy/analyses/base.py
```diff
@@ -337,8 +336,7 @@ def field(self) -> int | str:
 
         return "All" if field_number == 0 else field_number
 
-    @field.setter
-    def field(self, value: int | str):
+    def set_field(self, value: int | str):
```
crnh, line 339:
```suggestion
    def set_field(self, value: int | Literal["All"]):
```

zospy/analyses/base.py
```diff
@@ -324,9 +324,8 @@ def Results(self) -> _ZOSAPI.Analysis.Data.IAR_:  # noqa: N802
         """Analysis results."""
         return self._analysis.GetResults()
 
-    @property
-    def field(self) -> int | str:
-        """Gets the wavelength value of the analysis.
+    def get_field(self) -> int | str:
```
crnh, line 327:
```suggestion
    def get_field(self) -> int | Literal["All"]:
```

Furthermore, please put the `get_field` and `set_field` methods between the properties and `(get|set)_wavelength`.

crnh (changes requested):
Looks nice, thanks! I'd like to see some minor changes (see comments). Furthermore, are these methods never called from the unit tests?

LucVV:
I noticed that the `set_surface` for the new style analyses does not seem to have the fix that we implemented in the old version of `set_surface`, the current version only allows `'Objective'`, while some analyses might parse `'Object'` instead. I think this PR might be suitable to adres that, as it is also a setter.

LucVV:
> Furthermore, are these methods never called from the unit tests?
They are not directly called, but the tests use them under the hood. Maybe we should implement separate simple tests for them?

crnh:
> I noticed that the set_surface for the new style analyses does not seem to have the fix that we implemented in the old version of set_surface, the current version only allows 'Objective', while some analyses might parse 'Object' instead. I think this PR might be suitable to adres that, as it is also a setter.

Yes, good to apply that here as well!

> > Furthermore, are these methods never called from the unit tests?
> 
> They are not directly called, but the tests use them under the hood. Maybe we should implement separate simple tests for them?

In my opinion, it is indeed better to test them separately as well. The current tests may not cover all possible ways to call these methods. By testing them separately, we can ensure they are fully tested.

crnh:
> while some analyses might parse `'Object'` instead

By the way, if different analyses use different terms, it's probably best to standardize this, e.g. using only "Object", and maybe with a `SurfaceNumber` annotation, similar to the `WavelengthNumber` and `FieldNumber` annotations

LucVV:
I have made some changes, turned out that "Object" should not be used for `UseObjectiveSurface()` (and it is not really clear what this does. However, the API methods return messages that might contain error codes. I now utilize these to check if the passed value is accepted by the API (as this normally fails silently). I also implemented several unit tests for this.

I have one remaining question about which I'm not sure. If you pass an integer to `SetWavelengthNumber()`, 'SetFieldNumber()` or `SetSurfaceNumber()`, and the integer is higher than the number of wavelengths/fields/surfaces, the API reverts to the last wavelength/field/surface. Do we want to flag this or raise an error?

crnh (approved):

LucVV:
> > I have one remaining question about which I'm not sure. If you pass an integer to SetWavelengthNumber(), 'SetFieldNumber()orSetSurfaceNumber()`, and the integer is higher than the number of wavelengths/fields/surfaces, the API reverts to the last wavelength/field/surface. Do we want to flag this or raise an error?
> 
> You are getting something different than specified, so in my opinion we need to at least show a warning if this happens. Raising an exception would make sense as well (specifying a higher surface number than present in the system may indicate that something went wrong when building the system). 

I looked into this and had trouble obtaining e.g. the max surface number within `set_surface` as you do not have a refference to an `OpticStudioSystem` within `Analysis` (correct me if I'm wrong and overlooked something). This gives us several ways to implement that warning.
1. Have a reference to the `OpticStudioSystem` from which the analysis was created within the analysis. This might require several changes
2. Make an optional `max_surface` argument in `set_surface` that can be used to make the function less error-prone. We can than use that check in our analyses.
3. Perform these checks within dedicated analysis classes (e.g. `RayFan`) rather than in `set_surface`

crnh:
Let's do nothing for now. You can open an issue if you want, so we can solve it later on. Of course, it's always good to mention this behavior in the docstrings.

LucVV:
I added the warnings!

crnh (approved):

Issue #61, opened 2024-01-11
Deleting the `ZOS` instance doesn't (always) work


crnh:
### Problem

Several users of ZOSPy complained about being unable to create a new `ZOS` instance, even after deleting the existing instance using `del zos` (being unable to create a new instance when another instance exists is by design).
I looked into this and there are two main reasons for this problem:

1. The use of an interactive Python environment like IPython or Jupyter notebooks. These environments may keep some references to variables even after deleting them, so the existing `ZOS` instance is not removed by the garbage collector after calling `del zos`.
2. The existence of `OpticStudioSystem` instances. Every `OpticStudioSystem` contains a reference to the `ZOS` instance that created it. As a result, there will still be some references to this `ZOS` instance, and deleting `zos` will not result in its removal by the garbage collector. It is therefore necessary to 

I am not sure if this should be considered a bug, as this is related to how Python works. Furthermore, it is not necessary to delete and recreate `ZOS` objects, as the same instance can be used to manage all `OpticStudioSystem` instances. Nevertheless, users expect to be able to delete `ZOS` (probably due to the OpticStudio documentation suggesting to do so) so we should at least consider adding a note about this behaviour to the documentation. It may also be possible to automatically delete all `OpticStudioSystem` instances when deleting `zos`, but this most likely involves messing with the garbage collector, which may be dangerous.

### What version of ZOSPy are you running?

1.2.0

### What version of OpticStudio are you running?

23R2

### Which operating system do you use?

Windows 10 Enterprise

### In which environment do you use ZOSPy?

- [X] Plain Pythons scripts (not in Spyder)
- [X] Jupyter notebooks
- [ ] Spyder

### In which connection modes does the problem occur?

- [X] Extension mode
- [X] Standalone mode

### Example code

```Python
import logging

import zospy as zp

# Initiate logging so we can follow what ZOSPy is doing
logging.basicConfig(level=logging.DEBUG, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")

zos = zp.ZOS()
oss = zos.connect(mode="extension")

del zos

# See the output below: if this doesn't print the empty set, the instance was not deleted
print(f"ZOS instances: {zp.ZOS._instances}")
```

### Output

```Python
2024-01-11 12:46:33,917 - zospy.zpcore - DEBUG - Loading ZOS DLLs with preload set to False
[...]

PR #131, opened 2025-03-13
Prevent a second initialization of ZOS

crnh:
## Proposed change

`ZOS` is a singleton, meaning it can only be instantiated once. If a second instance is created, `ZOS.__new__` returns the existing instance. However, `__init__` is still called a second time on this already initialized `ZOS` instance. This causes some attributes, for example the connection to OpticStudio, to be reset to `None`. Unfortunately, this does not immediately break existing `OpticStudioSystem` instances, which is probably why we didn't notice this earlier.

I solved this by adding a new `__initialized` attribute, that is set to `True` when the `ZOS` instance is initialized. Initialization is skipped when this attribute is `True`.

This problem does not occur when using the decorator singleton pattern (see https://stackoverflow.com/questions/6760685/what-is-the-best-way-of-implementing-singleton-in-python). However, the decorator pattern breaks classmethods, and `ZOS` contains classmethods, so the current solution is probably the best we can achieve.

## Type of change

# Additional information

- Python version: 3.12
- OpticStudio version: N/A

## Checklist

### DIFF
--- a/tests/test_zpcore.py
+++ b/tests/test_zpcore.py
@@ -53,6 +53,16 @@ def test_zos_singleton(zos, oss):
         assert zos2 is zos
 
 
+@pytest.mark.filterwarnings("ignore:Only a single instance of ZOS can exist at any time")
+def test_zos_singleton_logs(zos, caplog):
+    with caplog.at_level("DEBUG"):
+        zos2 = zp.ZOS()
+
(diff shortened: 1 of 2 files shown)


jwmbeenakker (approved):

jwmbeenakker:
This one closes #61 ?

PR #128, opened 2025-03-10
Datagrid centration

LucVV:
## Proposed change

The way to interpret `MinX` and `MinY` from a datagrid seems to differ between analysis. In the current implementation, we label the columns and rows using:
```python
columns = np.linspace(datagrid.MinX, datagrid.MinX+ datagrid.Dx * (datagrid.Nx - 1), datagrid.Nx)
rows = np.linspace(datagrid.MinY, datagrid.MinY+ datagrid.Dy * (datagrid.Ny - 1), datagrid.Ny)
```

However, given [this discussion](https://community.zemax.com/zos-api-12/incorrect-datagrid-miny-for-geometric-image-analysis-5426) on the OpticStudio forum, it seems better center the labels on the center of the pixel by adjusting it to:
```python
minx = minx + 0.5 * datagrid.Dx
miny = miny + 0.5 * datagrid.Dy
columns = np.linspace(minx, minx + datagrid.Dx * (datagrid.Nx - 1), datagrid.Nx)
rows = np.linspace(miny, miny + datagrid.Dy * (datagrid.Ny - 1), datagrid.Ny)
```

When evaluating this implementation, it became apparent that `MinX` and `MinY` do however point to the center of the cell in some analyses, such as the `SurfaceCurvature` analysis, for which the old implementation should be kept. Thus, I added the keyword `cell_origin` to `unpack_datagrid` that can be used to specify how `MinX` and `MinY` should be treated (`cell_origin="bottom_left"` to set the origin to the bottom left of the cell, cell_origin="center" to set the origin to the center of the cell).

With these changes, the datagrids now seem to be nicely centered with appropriate labels. The only analysis where I encountered an additional problem is the `WavefrontMap`, where the datagrid did not span from `-1` to `1` as expected for a normalized analysis, but rather from `-1` to `1+1*datagrid.Dx`. I noticed furhtermore that the first row and column seems to be empty dummies. Thus, I now drop this row and column and then use `MinX` and `MinY` as origin for theadjusted datagrid. I have asked about this [on the OpticStudio forum](https://community.zemax.com/zos-api-12/incorrect-datagrid-miny-for-geometric-image-analysis-5426?postid=17856#post17856) to see if this handles it correctly.

As the columns and indices used in the dataframes that we return are slightly adjusted, this PR breaks many tests

## Type of change

- [ ] Example (a notebook demonstrating how to use ZOSPy for a specific application)
- [x] Bugfix (non-breaking change which fixes an issue)
[...]

PR #134, opened 2025-03-14
Keep old datagid label behaviour for analysis.old

LucVV:
## Proposed change

`ZOSPy` 2.0 has a slightly different way of labeling datagrid columns and indices, as implemented `unpack_datagrid` in #128. The analyses in `analyses.old` use the same `unpack_datagrid` as the new analyses and thus get the same labelling behaviour. As a result, the results of the current old analyses are not in line with the actual old analyses anymore. This also causes some unit tests for the old tests to fail.

While we do not really update the old analyses anymore, I think we should try to keep them equal. This requires two things:
1. Always passing `cell_origin="center"` to `unpack_datagrid`, as that essentially gives the old behaviour
2. Not rounding the column and indices labels in `unpack_datagrid`. To that end, I introduced the argument `label_rounding`, which specifies the decimals to which the labels are rounded. It defaults to 10 (the current rounding implementation), but als oaccepts None, which will result in no rounding. In the old analyses, `label_rounding=None` is now passed.

## Type of change

# Additional information

- Python version: 3.11
- OpticStudio version: 20.3.2

## Checklist

- [ ] I have followed the [contribution guidelines][contribution-guidelines]
- [ ] The code has been linted, formatted and tested locally using hatch.

### DIFF
--- a/zospy/utils/zputils.py
+++ b/zospy/utils/zputils.py
@@ -91,9 +91,10 @@ def unpack_dataseries(dataseries: _ZOSAPI.Analysis.Data.IAR_DataSeries) -> pd.Da
 
 def unpack_datagrid(
     datagrid: _ZOSAPI.Analysis.Data.IAR_DataGrid,
-    minx=None,
-    miny=None,
+    minx: float | None = None,
+    miny: float | None = None,
     cell_origin: Literal["bottom_left", "center"] = "bottom_left",
+    label_rounding: int | None = 10,
 ) -> pd.DataFrame:
     """Unpack an OpticStudio datagrid to a Pandas DataFrame.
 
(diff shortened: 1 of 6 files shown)


zospy/utils/zputils.py
```diff
@@ -109,6 +110,9 @@ def unpack_datagrid(
         Defines how minx and miny are handled to determine coordinates. Either 'bottom_left' indicating that they are
         defining the bottom left of the grd cell, or 'center', indicating that they provide the center of the grid cell.
         Defaults to 'bottom_left'.
+    label_rounding : Optional[int]
```
crnh, line 113:
```suggestion
    label_rounding : int, optional
```

crnh:
Thanks for noticing and fixing this! Tests are currently running.

crnh:
Tests pass. Thanks a lot!

PR #154, opened 2025-05-22
Set workflow permissions

crnh:
Set workflow permissions as advised by dependabot.


crnh:
I tested the test publish workflow and this workflow runs successfully.

PR #225, opened 2026-06-03
Normunpol batch raytrace

andibarg:
### 1. NormUnpol Batch Raytrace

I added the NormUnpol batch raytrace as a new analysis under `raysandspots` - hope this is the right place for it. It can be used, for instance, to create spot diagrams (see the official examples). 

The batch ray trace is technically not an OpticsStudio analysis but I thought we could perhaps still treat it as such. Note that I could not add the settings to the RayTrace analysis under `run_analysis` as in `self.analysis.Settings.Hx = self.settings.hx` because it does not accept lists. Perhaps there is a workaround or better solution to this.

Unfortunately, I cannot run/provide Jupyter notebook examples at the moment. Instead, I prepared this MWE:
```python
import numpy as np
from pathlib import Path
import zospy as zp

zos = zp.ZOS()
oss = zos.connect(mode="standalone")

system_file = Path(zos.Application.SamplesDir) / "Sequential/Objectives/Double Gauss 28 degree field.zmx"
oss.load(system_file)

N = 3000
df = zp.analyses.raysandspots.BatchRayTraceNormUnpol(
    hx=np.linspace(-1,1,N),
    hy=np.zeros(N),
    px=np.zeros(N),
    py=np.zeros(N),
    wavelength=1,
    surface="Image",
).run(oss).data

print(df)
``` 

### 2. Removed warning from SingleRayTrace when surface comment is missing

There is a warning when running the single ray trace in the Double Gauss example. This is simply because some surfaces in the .zmx file do not have a comment. Even though I understand the reason for the warning, I thought it would be nice to avoid the warning in this case.

## Type of change

- [x] New analysis (a wrapper around an OpticStudio analysis)

# Additional information

- OpticStudio version: 2024 R2.01 Premium

## Checklist

- [x] The code has been linted, formatted and tested (see the [contribution guidelines][code-style-guidelines]).
- [ ] I tested ZOSPy for _all_ supported Python versions (`hatch test -a`).

If you updated an example:

- [ ] I executed all examples and verified that they run without errors (`hatch run all-examples`).

[contribution-guidelines]: https://zospy.readthedocs.io/en/latest/contributing/contributing.html
[code-style-guidelines]: https://zospy.readthedocs.io/en/latest/contributing/contributing.html#workflow
[unittest-instructions]: https://zospy.readthedocs.io/en/latest/contributing/unit_tests.html


andibarg:
Btw, sorry if I am not following procedure very well for the pull request. I am not too familiar with it yet.

zospy/analyses/raysandspots/__init__.py
```diff
@@ -7,5 +7,8 @@
     SingleRayTrace,
     SingleRayTraceSettings,
 )
+from zospy.analyses.raysandspots.batch_ray_trace import NormUnpol as BatchRayTraceNormUnpol
+from zospy.analyses.raysandspots.batch_ray_trace import NormUnpolSettings as BatchRayTraceNormUnpolSettings
```
crnh: Do not use alias imports, but rename the classes to `BatchRayTraceNormUnpol` and `BatchRayTraceNormUnpolSettings` instead.

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+
+from zospy.analyses.base import BaseAnalysisWrapper
+from zospy.analyses.decorators import analysis_settings
+from zospy.api import constants
+
+__all__ = ("NormUnpol", "NormUnpolSettings")
+
+
+# Constrained types
+FloatInRange = Annotated[float, confloat(le=1.0, ge=-1.0)]
```
crnh: `FloatInRange` is not a very clear name, since it does not specify the range. You could consider renaming it to `NormalizedCoordinate`, which does imply a range of [-1, 1].

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+        Type of rays to trace ('Real' or 'Paraxial'). Defaults to 'Real'.
+    opd_mode : zospy.api.constants.Tools.RayTrace.OPDMode
+        Mode of optical path difference for rays (e.g. 'None'). Defaults to 'None'.
+    """
+
+    hx: list[FloatInRange] = Field(default=[0], description="Normalized X field coordinate")
+    hy: list[FloatInRange] = Field(default=[0], description="Normalized Y field coordinate")
+    px: list[FloatInRange] = Field(default=[0], description="Normalized X pupil coordinate")
+    py: list[FloatInRange] = Field(default=[0], description="Normalized Y pupil coordinate")
+    wavelength: Union[PositiveInt, list[PositiveInt]] = Field(default=1, description="Wavelength number")
```
crnh:
```suggestion
    wavelength: PositiveInt | Sequence[PositiveInt] = Field(default=1, description="Wavelength number")
```

crnh: Use `Sequence` instead of `list`, so tuples and ranges can be used as well

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+    surface: Literal["Image"] | Annotated[int, Field(ge=0)] = Field(default="Image", description="Surface number")
+    rays_type: Literal["Real", "Paraxial"] = Field(default="Real", description="Type of rays to trace")
+    opd_mode: Literal["None", "Current", "CurrentAndChief"] = Field(default="None", description="Mode of optical path difference")
```
crnh: For ZOSAPI constants, we have the `ZOSAPIConstant` annotation in `zospy.analyses.parsers.types`:

https://github.com/MREYE-LUMC/ZOSPy/blob/79ce80c7392c536c2b1f2855598ed0ccf96cc91f/zospy/analyses/parsers/types.py#L156

Please use this instead of a literal, so ZOSAPI / ZOSPy constants can be passed as well. For an example, see the FFT PSF analysis:

https://github.com/MREYE-LUMC/ZOSPy/blob/79ce80c7392c536c2b1f2855598ed0ccf96cc91f/zospy/analyses/psf/fft_psf.py#L57

<sub>I see the Ray Trace analysis still uses a literal where it could use a constant; that is unfortunate, so I will fix that soon.</sub>

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @model_validator(mode="after")
+    def validate_lengths(self):
+        """Validate that hx, hy, px, py (and wavelength) have the same lengths."""
+        lengths = [len(self.hx), len(self.hy), len(self.px), len(self.py)]
+        if len(set(lengths)) != 1:
```
crnh: Just check for `not len(self.hx) == len(self.hy) == len(self.px) == len(self.py)`, which is more explicit, and do not include the lengths in the exception message.

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+            raise ValueError(
+                f"Hx, Hy, Px, Py must all have the same length; got lengths {lengths}."
+            )
+        expected_len = lengths[0]
+
+        if isinstance(self.wavelength, list):
```
crnh:
```suggestion
        if isinstance(self.wavelength, Sequence):
```

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,155 @@
...
+                raise ValueError(
+                    f"`wavelength` list length ({len(self.wavelength)}) must match "
+                    f"the length of Hx, Hy, Px, Py ({expected_len})."
+                )
+
+        return self
+    
+
+class NormUnpol(BaseAnalysisWrapper[pd.DataFrame | None, NormUnpolSettings],
+    analysis_type="RayTrace",
```
crnh: This causes an unused `RayTrace` analysis to be created before running the ray trace tool. Just don't specify an analysis type and override `BaseAnalysisWrapper._create_analysis()` with an empty method. I expect this should be sufficient to bypass analysis creation for this class.

If this doesn't work, please let us know so we can work on a better solution.

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+):
+    """Batch Ray Trace of unpolarized light."""
+
+    def __init__(
+        self,
+        *,
+        hx: list[float] = [0],
+        hy: list[float] = [0],
+        px: list[float] = [0],
+        py: list[float] = [0],
```
crnh: Do not use mutable default arguments. `Ruff` should will complain about this as well. Instead, use Sequences with a tuple as default argument.


```suggestion
        hx: Sequence[float] = (0,),
        hy: Sequence[float] = (0,),
        px: Sequence[float] = (0,),
        py: Sequence[float] = (0,),
```

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+        wavelength: int | list[int] = 1,
+        surface: Literal["Image"] | int = "Image",
+        rays_type: str = "Real",
+        opd_mode: str = "None",
```
crnh: Annotate with a union of `str` and a constant type, like `psf_type` in the FFT PSF analysis:

https://github.com/MREYE-LUMC/ZOSPy/blob/79ce80c7392c536c2b1f2855598ed0ccf96cc91f/zospy/analyses/psf/fft_psf.py#L75

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+    ):
+        """Create a new Batch Ray Trace of unpolarized light.
+
+        See Also
+        --------
+        NormUnpolSettings : Settings for the Batch Ray Trace of unpolarized light.
+        """
+        super().__init__(settings_kws=locals())
+
+    def run_analysis(self) -> pd.DataFrame | None:
```
crnh: Remove `None` from the return types. If anything fails, the analysis should raise an exception, not return `None`.

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+            if len(self.wavelength) != expected_len:
+                raise ValueError(
+                    f"`wavelength` list length ({len(self.wavelength)}) must match "
+                    f"the length of Hx, Hy, Px, Py ({expected_len})."
+                )
+
+        return self
+    
+
+class NormUnpol(BaseAnalysisWrapper[pd.DataFrame | None, NormUnpolSettings],
```
crnh:
```suggestion
class NormUnpol(BaseAnalysisWrapper[pd.DataFrame, NormUnpolSettings],
```

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        super().__init__(settings_kws=locals())
+
+    def run_analysis(self) -> pd.DataFrame | None:
+        """Run the Batch Ray Trace of unpolarized light."""
+        number_of_rays = len(self.settings.hx)
+        if isinstance(self.settings.wavelength, int):
+            wvls = [self.settings.wavelength]*number_of_rays
+        else:
+            wvls = self.settings.wavelength
```
crnh: Only use abbreviations if there is a good reason to do so, and use a ternary operator here:

```suggestion
        wavelengths = [self.settings.wavelength] * number_of_rays if isinstance(self.settings.wavelength, int) else self.settings.wavelength
```

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+        # Add rays
+        for ii in range(number_of_rays):
+            normUnPolData.AddRay(
+                wvls[ii],
+                self.settings.hx[ii],
+                self.settings.hy[ii],
+                self.settings.px[ii],
+                self.settings.py[ii],
+                opd_mode,
+            )
```
crnh: In general, use `i` as the iteration variable instead of `ii`.
In this specific case, `zip` the sequences together:


```suggestion
        for wavelength, hx, hy, px, py in zip(
            wavelengths,
            self.settings.hx,
            self.settings.hy,
            self.settings.px,
            self.settings.py,
        ):
            normUnPolData.AddRay(wavelength, hx, hy, px, py, opd_mode)
```

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,155 @@
...
+                wvls[ii],
+                self.settings.hx[ii],
+                self.settings.hy[ii],
+                self.settings.px[ii],
+                self.settings.py[ii],
+                opd_mode,
+            )
+
+        # Run ray trace and read results
+        self.analysis.ApplyAndWaitForCompletion()
```
crnh: Why do you do this? Is the output of the raytrace analysis used anywhere?

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+from zospy.analyses.base import BaseAnalysisWrapper
+from zospy.analyses.decorators import analysis_settings
+from zospy.api import constants
+
+__all__ = ("NormUnpol", "NormUnpolSettings")
+
+
+# Constrained types
+FloatInRange = Annotated[float, confloat(le=1.0, ge=-1.0)]
+PositiveInt = Annotated[int, conint(gt=0)]
```
crnh: Use Pydantic's built-in `PositiveInt` type instead.

crnh: `confloat` and `conint` are deprecated: https://pydantic.dev/docs/validation/latest/api/pydantic/types/#pydantic.types.confloat. Moreover, they return a type and therefore don't work with `Annotated` types. Instead, use an annotation with `Field`:

`FloatInRange = Annotated[float, Field(le=1.0, ge=-1.0)]`

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+                self.settings.hy[ii],
+                self.settings.px[ii],
+                self.settings.py[ii],
+                opd_mode,
+            )
+
+        # Run ray trace and read results
+        self.analysis.ApplyAndWaitForCompletion()
+        raytrace.RunAndWaitForCompletion()
+        normUnPolData.StartReadingResults()
```
crnh: Use snake_case variable names

zospy/analyses/raysandspots/batch_ray_trace.py
```diff
@@ -0,0 +1,155 @@
...
+        # Create placeholders for all output arguments
+        placeholders = [Int32(1)]*3 + [Double(1.0)]*11
+
+        # Read all results and append to outputs
+        outputs = []
+        columns = ["rayNumber","ErrorCode","vignetteCode",
+                   "X","Y","Z","L","M","N",
+                   "l2","m2","n2","opd","Intensity"]
+        for ii in range(number_of_rays):
+            outputs.append(normUnPolData.ReadNextResult(*placeholders)[1:])
```
crnh: Python.NET doesn't require specifying the `out` parameters. Instead, you can generate the output list with a comprehension:

```python
outputs = [normUnPolData.ReadNextResult()[1:] for _ in range(number_of_rays)
```

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,155 @@
...
+        rays_type = constants.process_constant(
+            constants.Tools.RayTrace.RaysType, self.settings.rays_type
+        )
+        opd_mode = constants.process_constant(
+            constants.Tools.RayTrace.OPDMode, self.settings.opd_mode
+        )
+        surface = -1 if self.settings.surface == "Image" else self.settings.surface
+
+        # Initiate batch ray trace
+        raytrace = self.oss.Tools.OpenBatchRayTrace()
```
crnh: If a different tool is open, this analysis will fail. The analysis should first check for other open tools before opening the batch ray race tool.

crnh (changes requested):
Very nice work, thanks! I agree that `zospy.analyses.raysandspots` is the best place for this feature, although it is strictly spoken not an analysis.

> Btw, sorry if I am not following procedure very well for the pull request. I am not too familiar with it yet.

No problem, we are happy to guide you through the process!

A few general comments:

- Read the [contributing guidelines](https://zospy.readthedocs.io/en/latest/contributing/contributing.html) for instructions on how to lint, format and test ZOSPy. Several points in this review could have been discovered and fixed by using a linter. Furthermore, the linting and formatting checks are required to pass in order to merge the PR.
- For me, this analysis only works in standalone mode. In extension mode, I get a `RemotingException`. Is this expected? If so, the analysis should check for the connection mode and raise a clear exception in extension mode.
- Tests are still missing. The issues with settings validation (where px=[2] is outside the allowed range, but still accepted) could have been caught earlier by adding tests.
- It would be better to manage the batch ray trace tool with a context manager, to ensure the tool will be closed even after an exception. We can consider providing a utility function for this in a separate PR.

I agree the warning about the comment in the raytrace analysis is annoying; I always suppress it in my code, but thanks for addressing this. Since this is a separate change, it belongs in a separate PR. This time we'll accept this, but next time please open separate PRs for different changes.

zospy/tools/batch_raytrace.py
andibarg: The analysis is not used anywhere. This is a workaround because `BaseAnalysisWrapper` needs the metadata that `self.analysis.ApplyAndWaitForCompletion()` seem to generate. Without this line I get an error in base.py, line 286: `AttributeError: 'IA_' object has no attribute 'metadata'`.

zospy/analyses/raysandspots/__init__.py
andibarg: Done.

zospy/analyses/raysandspots/batch_ray_trace.py
andibarg: Done

zospy/analyses/raysandspots/batch_ray_trace.py
andibarg: Done

zospy/analyses/raysandspots/batch_ray_trace.py
andibarg: Done

zospy/tools/batch_raytrace.py
crnh: Okay, I see. In my opinion we shouldn't want to run an analysis just to get a tool working in a `BaseAnalysisWrapper` class. This indicates the current implementation as an analysis is unfortunately not the right approach. 

Instead, I propose a different approach also follows the ZOS-API semantics more closely:

1. Create a new `zospy.tools` submodule;
2. Create a `BaseToolWrapper` base class. This class can be modeled after the `BaseAnalysisWrapper` class, but it can be much simpler, because it doesn't need to manage text files, config files, completion handling etc.
3. Implement the batch ray trace using the new `BaseToolWrapper` class.

If you want, I can implement 1 and 2.

zospy/analyses/raysandspots/batch_ray_trace.py
andibarg: Done. I guess for the surface `Literal["Image"]` is ok, since I found it in many other analyses.

zospy/tools/batch_raytrace.py
andibarg: I tried removing the analysis type and adding `def _create_analysis(self): pass` but I get `AttributeError: 'NoneType' object has no attribute 'metadata'`. It seems that `BaseAnalysisWrapper` needs `analysis` and `metadata` to function properly (see also my comment below about using `self.analysis.ApplyAndWaitForCompletion()` as a workaround).

zospy/tools/batch_raytrace.py
andibarg: Sorry, I am not quite sure how to do this. Could you please help me out here?

andibarg:
Thanks a lot for your review! 

I tried implementing your points as well as I could. 

Just a comment about a change: I am not very happy with using `Sequence` instead of `list` because it bans numpy arrays (see my MWE above). But I cannot find a better solution at the moment.

Btw, I think it is expected that the batch raytrace does not work in extension mode because it is not really an analysis that can be run by itself from within OpticStudio.

zospy/tools/batch_raytrace.py
crnh: You can get the current tool with `oss.Tools.CurrentTool`. If no tool is opened, this will be `None`. However, I am implementing a context manager for safe management of tools in #226, which also performs this check. You can wait until that one gets merged and use it in the batch ray trace analysis.

zospy/tools/batch_raytrace.py
crnh: Yes, this is a clear indication that the current approach is not the right approach. See https://github.com/MREYE-LUMC/ZOSPy/pull/225#discussion_r3362052119 for an alternative.

zospy/tools/batch_raytrace.py
andibarg: Sounds reasonable to me.

I hope if `BaseToolWrapper` and the new decorator settings (`tools_settings`?) are similar to the ones in `zospy.analyses` then the batch ray trace code does not even need too many changes :).

crnh:
> Just a comment about a change: I am not very happy with using Sequence instead of list because it bans numpy arrays (see my MWE above). But I cannot find a better solution at the moment.

Yes, I noticed that as well. The best way to allow NumPy arrays as well is probably typing these attributes with a union of Sequence and NDArray. This requires a custom validator to ensure additional constraints are applied to the array items as well.

zospy/analyses/raysandspots/batch_ray_trace.py
crnh: > I guess for the surface Literal["Image"] is ok, since I found it in many other analyses.

Yes, constant handling for the surface is implemented in the `Analysis` class.

zospy/tools/batch_raytrace.py
crnh: They will be very similar; I think I will reuse `analysis_settings` for the settings. Migrating your current implementation to the new one should be straightforward. I am currently implementing this in #226.

Wellidontcare:
Something that could be cool to implement in the future (not a feature request):

Zemax batch tracing, needs you to add the rays one by one, this is a huge slowdown due to python loops and API overhead, retrieval of traced rays suffers from this as well. 

There is the option to outsource the ray adding and retreival to C#: https://optics.ansys.com/hc/en-us/articles/42661765866899-Batch-Processing-of-Ray-Trace-Data-using-ZOS-API-in-MATLAB-or-Python (the example is bugged by the way, but the idea is there)

With this extra wrapper you could pass and retrieve the rays via numpy arrays. If this would be implemented in ZOSPy, the project would need to ship the C# DLL though (question is, where to put it, as git does not like binaries), cross platform should not be an issue, as Zemax is Windows only anyways.

andibarg:
I managed to transfer the batch raytrace over to `zospy.tools` and to me this looks good now. Please let me know what you think.

Regarding unit tests, I am having difficulties running `hatch` correctly. I think this might be for the same reason why I cannot run/provide Jupyter notebook examples (long story). I wanted to update the Double Gauss example with my MWE but I am afraid I am a bit stuck at the moment.

@Wellidontcare: Sounds interesting. For my use cases, the python loop does not really cause a huge slowdown but I can imagine there are applications where this matters. Would be cool to implement in the future.

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,170 @@
...
+        Sequence of normalized X field coordinates. Defaults to [0].
+    hy : Sequence[float]
+        Sequence of normalized Y field coordinates. Defaults to [0].
+    px : Sequence[float]
+        Sequence of normalized X pupil coordinates. Defaults to [0].
+    py : Sequence[float]
+        Sequence of normalized Y pupil coordinates. Defaults to [0].
+    wavelength : int | Sequence[int]
+        The wavelength number that is to be used. Must be an integer or a sequence of integers specifying the wavelength number.
+        Defaults to 1.
```
crnh: Please add ndarrays to the types in the docstring

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,170 @@
...
+    rays_type : zospy.api.constants.Tools.RayTrace.RaysType
+        Type of rays to trace ('Real' or 'Paraxial'). Defaults to 'Real'.
+    opd_mode : zospy.api.constants.Tools.RayTrace.OPDMode
+        Mode of optical path difference for rays (e.g. 'None'). Defaults to 'None'.
+    """
+
+    hx: CoordinateVector = Field(default=[0], description="Normalized X field coordinate")
+    hy: CoordinateVector = Field(default=[0], description="Normalized Y field coordinate")
+    px: CoordinateVector = Field(default=[0], description="Normalized X pupil coordinate")
+    py: CoordinateVector = Field(default=[0], description="Normalized Y pupil coordinate")
```
crnh: Pydantic supports mutable default values, but to keep things in sync with the `BatchRayTraceNormUnpol` constructor's signature it's better to use tuples here as well.

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,170 @@
...
+        tool.Py = self.settings.py
+        tool.number_of_rays = len(tool.Hx)
+        tool.wavelengths = [self.settings.wavelength] * number_of_rays if isinstance(self.settings.wavelength, int) else self.settings.wavelength
+        tool.rays_type = constants.process_constant(
+            constants.Tools.RayTrace.RaysType, self.settings.rays_type
+        )
+        tool.opd_mode = constants.process_constant(
+            constants.Tools.RayTrace.OPDMode, self.settings.opd_mode
+        )
+        tool.surface = -1 if self.settings.surface == "Image" else self.settings.surface
```
crnh: I still have to test is, but are you sure the ZOS-API uses snake_case (`rays_type`) instead of PascalCase (`RaysType`) names for the number of rays, wavelengths, rays type, opd mode and surface? Please double-check this. Setting an incorrectly named attribute will monkeypatch the `tool` object instead of raising an exception, and OpticStudio will silently use the default value.

zospy/tools/batch_raytrace.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Read all results and append to outputs
+        outputs = []
+        columns = ["rayNumber","ErrorCode","vignetteCode",
+                   "X","Y","Z","L","M","N",
+                   "l2","m2","n2","opd","Intensity"]
+        outputs = [norm_unpol_data.ReadNextResult()[1:] for _ in range(tool.number_of_rays)]
+
+        # Convert to DataFrame and return
+        return pd.DataFrame(outputs, columns=columns)
```
crnh, line 193: Please document the structure of the `DataFrame`  in the docstring of this class.

crnh (commented):
Thanks for updating this! Due to a conference, the final tests before merging this PR will probably happen next week.

This PR should be merged _after_ #226, because it depends on features that are introduced there.

crnh:
> Regarding unit tests, I am having difficulties running `hatch` correctly. I think this might be for the same reason why I cannot run/provide Jupyter notebook examples (long story).

Can you let us know what goes wrong, so we can maybe help you and if applicable update our contribution guidelines? If you don't want to discuss this as part of this PR, please drop us a message at [zospy@mreye.nl](mailto:zospy@mreye.nl). 

If the issue is related to `hatch`, you can always run the tests with `pytest`. `hatch` primarily makes setting up the test environment easier.

crnh:
@andibarg I just merged #226. Please complete these steps to finalize this PR:

- [ ] Rebase this branch onto main
- [ ] Address the open review comments
- [ ] Add unit tests for the batch ray trace tool

zospy/tools/batch_raytrace.py
andibarg: I am not sure about this. In the `ZemaxUI.ZOSAPI.Tools.BatchRayTraceTool` object (from `oss.Tools.CurrentTool`), the only relevant attributes I can see are `MaxRays`, `SurfaceNumber`, and `Wavelength` (so yes, PascalCase). All of them are integers so I cannot set the `wavelengths` array, for example.

Is it necessary to set any attributes of `tool` here? You see what I mean in my latest commit. What do you think?

crnh: Looks like it is indeed not necessary to set the attributes, as they are all set with `CreateNormUnpol` and `AddRay`. The current solution looks good to me, although I still have to test if this works.

crnh:
@andibarg I added unit tests for the batch ray trace tool and updated the changelog. Once the tests pass for all supported Python versions, this one is good to go!

crnh (dismissed):

crnh (approved):

